PR #420, opened 2025-03-10
Pre-commit configuration for Go and Markdown formatting

pdyraga:
References: https://linear.app/thesis-co/issue/TET-87/contributing-policy-contributingmd

### Introduction

As a part of the work on the contributing policy, I realized we do not have a pre-commit configuration in the repository. This pull request adds pre-commit configuration for Go and Markdown file formatting.

### Changes

I wanted to "fix everything" (TM) in this PR but quickly realized there is a disconnect between what GitHub Actions expect, what we have in the Makefile, and what we have in the code.

For example, if you run `make lint`, you will see the linter reports an error in one of the files. This error was not caught by GitHub actions when this code was pushed. Another example is the use of `cosmos/gosec` scanner in GitHub actions - this scanner is no longer maintained, and it is unclear whether we can have it supported from pre-commit. Similar story about the superlinter config from GitHub actions

Thus, baby-steps. Let's fix the formatting for the files we edit most often and then go after the rest.

### Testing

Insert some changes in one of the Go files the formatters is not going to like. Add this file to the index and run `git commit`. The pre-commit rules should block you. Then run `make lint-fix` and add the updated file to the index. Now, `git commit` should work. Try the same with a Markdown file and `make format-markdown-fix`.

---

### Author's checklist

### Reviewer's checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,3 +13,9 @@ repos:
         files: '\.go$'
         language: script
         description: "Format Go code"
+      - id: markdown-format
+        name: 'format markdown'
+        entry: /usr/bin/env bash -c 'make format-markdown'
+        files: '\.md'
+        language: script
+        description: "Format Markdown files"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,3 +3,13 @@ repos:
     rev: 31903eabdb97f5a3375fb1ea419ff782de64d7e7
     hooks:
       - id: check-added-large-files
+  - repo: local
+    hooks:
+      - id: go-format
+        name: 'format go'
+        # gofumpt does not return error exit code now, so we need to workaround
+        # it with `test -z`. See https://github.com/mvdan/gofumpt/issues/114.
+        entry: /usr/bin/env bash -c 'test -z "$(make -s format)"'
+        files: '\.go$'
+        language: script
+        description: "Format Go code"
--- a/Makefile
+++ b/Makefile
@@ -234,6 +234,12 @@ format-fix:
 
 .PHONY: format format-fix
 
+format-markdown:
+	markdownlint '**/*.md' --config .markdownlint.yml
+
+format-markdown-fix:
+	markdownlint '**/*.md' --config .markdownlint.yml --fix
+
 ###############################################################################
 ###                                Protobuf                                 ###
 ###############################################################################
--- a/Makefile
+++ b/Makefile
@@ -216,10 +216,23 @@ lint-fix:
 
 .PHONY: lint lint-fix
 
-format:
-	find . -name '*.go' -type f -not -path "./vendor*" -not -path "*.git*" -not -path "./client/docs/statik/statik.go" -not -name '*.pb.go' | xargs gofumpt -w -l
+format: 
+	find . -name "*.go" -type f \
+	  -not -path "./vendor*" \
+	  -not -path "*.git*" \
+	  -not -path "./client/docs/statik/statik.go" \
+	  -not -name "*.pb.go" \
+	  -not -name "*.pb.gw.go" | xargs gofumpt -l
 
-.PHONY: format
+format-fix:
+	find . -name "*.go" -type f \
+	  -not -path "./vendor*" \
+	  -not -path "*.git*" \
+	  -not -path "./client/docs/statik/statik.go" \
+	  -not -name "*.pb.go" \
+	  -not -name "*.pb.gw.go" | xargs gofumpt -w -l
+
+.PHONY: format format-fix
 
 ###############################################################################
 ###                                Protobuf                                 ###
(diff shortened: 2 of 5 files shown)


jeremyletang (commented):
The three go file updated in the PR are codegen, shouldn't they be excluded from the linting?

Also, you may want to request review from: mezo-org/mezod-developers instead of mezo-org/mezodrome-developers

jeremyletang:
> Good call. I am looking at the `gfumpt` excluded paths in the Makefile now, and I think we should be able to get rid of some rules from there as well.

I run the linter using golangci-lint. Running the following at the root of the repo should have the expected behaviour against all the expected file + fixe the issues if possible (including gofumpt):

```
golangci-lint run -v --fix --config .golangci.yml
```

.pre-commit-config.yaml
```diff
@@ -3,3 +3,19 @@ repos:
...
+  - repo: local
+    hooks:
+      - id: go-format
+        name: 'format go'
+        # gofumpt does not return error exit code now, so we need to workaround
+        # it with `test -z`. See https://github.com/mvdan/gofumpt/issues/114.
+        entry: /usr/bin/env bash -c 'test -z "$(make -s format)"'
+        files: '\.go$'
+        language: script
+        description: "Format Go code"
```
lukasz-zimnoch, line 15: I don't think I ever used `make format` during the work on `mezod`. The `make lint-fix` is the default way to go as this is used in CI as well (see lint.yml). Shouldn't we remove `format` altogether and use `lint`?

pdyraga: The conversation we have is a good example of why we need to fix it :) Historically, we have been separating linting from formatting, so seeing both `lint` and `format` targets in the Makefile, I was assuming that's what we do in this repository as well. 

I am down using just `lint` target and removing `format` and `format-fix` altogether. I just tested locally, and `make lint` detects the formatting problems. I would also add `make lint` to the pre-commit config and leave `format-markdown` and `format-markdown-fix` for now.

Outside of this PR, we need to investigate what's up with the CI - it should break given the linter complains with:
```
❯ make lint
golangci-lint run --out-format=tab
x/bridge/abci/proposal.go:554:24      gosec    G601: Implicit memory aliasing in for loop.
make: *** [lint] Error 1
```

But that's something that can wait given the pre-commit should start complaining after the proposed changes.

How does it sound?

pdyraga: Done in f068bc8765f8b848f91732842aa79350feb23e7c. Also, the PR description is updated to reflect the new reality. The G601 error was fixed separately in #422. I did not want to mix pre-commit config changes and Go source code changes in one PR.

lukasz-zimnoch: > I am down using just lint target and removing format and format-fix altogether. I just tested locally, and make lint detects the formatting problems. I would also add make lint to the pre-commit config and leave format-markdown and format-markdown-fix for now.

Makes sense!

> Outside of this PR, we need to investigate what's up with the CI - it should break given the linter complains with:

CI goes over diffs only, see [this](https://github.com/mezo-org/mezod/blob/9b32724957a0a1b7a8a330d622e97e259db375e9/.github/workflows/lint.yml#L36). Hence why there is no failure.

pdyraga: It is not failing now, but it should have failed when we committed that problematic change, no?

jeremyletang: Just by curiosity @pdyraga what version of golangci-lint are you using? I'm on:
```
➜  mezod git:(collecting-fees) golangci-lint --version
golangci-lint has version v1.60.1 built with go1.23.4 from (unknown, modified: ?, mod sum: "h1:DRKNqNTQRLBJZ1il5u4fvgLQCjQc7QFs0DbhksJtVJE=") on (unknown)
```

And running `make lint` returns no error for me. if it's a version older than compiled with 1.22, that might be the reason you are getting the error / the CI not because the behaviour the linter listed have changed since 1.22 (https://go.dev/blog/loopvar-preview). Now that doesn't mean we shouldn't prevent it, but it should be fine with any recent go compiler (also, it's be a problem with go routines).

Anyway, my point here is not really about this specific problem, but we may want to enforce a specific version for these tools as well if not done already.

lukasz-zimnoch: Maybe we used a different version of golangci-lint that did not detect this back then? Hard to say now.

jeremyletang: > Maybe we used a different version of golangci-lint that did not detect this back then? Hard to say now.

I believe you are right @lukasz-zimnoch

pdyraga: I am on an older version, yes:
```
❯ golangci-lint --version
golangci-lint has version 1.55.2 built with go1.21.3 from e3c2265 on 2023-11-02T21:40:02Z
```

Good idea about enforcing the version, I'll add it to #421 .

PR #422, opened 2025-03-12
Avoid memory aliasing when preparing bridging proposal

pdyraga:
### Introduction

This change removes the potential issue with memory aliasing. The `event` is a loop variable so even though its value changes, the memory location does not. As a result, we've been passing the same memory location to the `addVote` function. This worked fine because we were not capturing this memory location anywhere and were just extracting values from the pointer but to make this code more future-proof we should pass the address of the actual element from the `AssetsLockedEvents` slice.

### Changes

This is a one-linter technical change in `proposal.go`.

### Testing

No testing is required beyond checking that all tests pass. You can also confirm `make lint` no longer complains.

---

### Author's checklist

### Reviewer's checklist

### DIFF
--- a/x/bridge/abci/proposal.go
+++ b/x/bridge/abci/proposal.go
@@ -550,8 +550,8 @@ func (ale *assetsLockedExtractor) CanonicalEvents(
 		// addVote assumes the given event is valid and does not perform
 		// any validation over it. We ensure validity by calling
 		// validateAssetsLockedEvents above.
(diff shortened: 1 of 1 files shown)


jeremyletang (approved):

lukasz-zimnoch (commented):
Looks good!

lukasz-zimnoch (approved):

PR #436, opened 2025-04-02
Support mainnet `MezoBridge` contract

lukasz-zimnoch:
Closes: https://linear.app/thesis-co/issue/TET-62/mainnet-binding-for-the-mezobridge-contract

### Introduction

Here we add support for the `MezoBridge` contract deployed on Ethereum Mainnet.

#### Network-specific Ethereum bindings

So far, Ethereum bindings for Mezo Portal were network-agnostic and were living inside `portal/gen`. Those bindings were populated for a specific environment using `make bindings` so the following `make build` resulting in an environment-specific binary. This approach is not flexible as it forces us to maintain separate `mezod` binaries for testnet  and mainnet.

Here we change that and add simultaneous support of multiple Ethereum networks in the bindings directory structure. Now, `make bindings` create separate binding directory for each supported Ethereum network: `portal/sepolia/gen` and `portal/mainnet/gen`. 

#### Checkout contract addresses to VCS

As Ethereum bindings support multiple networks now, contract addresses don't have to be dynamic and can be checked out into the version control system. This way, `make bindings` does no longer have to be run before `make build` public binaries.  From now on, this command has to be executed only if contracts ABI or addresses changed and needs to be included in the `mezod` binary.

#### Ethereum bindings facade

The new network-specific Ethereum bindings must be consumed by client code (e.g. Ethereum sidecar server). Loading bindings based on the underlying Ethereum network at runtime is quite challenging. Moreover, it's not really needed as for now, Mezo Portal contracts have the same interfaces on Ethereum Mainnet and Sepolia. Basically, we just need to load the network-specific addresses.

Because of that, we are adding the new `portal` package acting as a facade for the bindings and just pin up the Mainnet bindings under it while providing network-specific contract address resolution. This approach is sustainable as long as there are no differences between Mainnet and Sepolia contracts. Once such differences emerge, we need to revisit it.

#### Changes in the Ethereum sidecar

So far, the Ethereum sidecar server directly depended on the contract bindings. Here we adjust it to depend on the new `portal` facade package. This decouples the sidecar from the bindings mechanism. In the future, all complexity related to different binding environments can be seamlessly handled inside the `portal` facade package.

### Testing

Run the Ethereum sidecar using `mezod ethereum-sidecar` against different Ethereum networks. This can be set using the `--ethereum-sidecar.server.network` flag (values are `sepolia` or `mainnet`). Confirm the sidecar properly resolves the network and observes the proper `MezoBridge` contract (addresses can be found [here](https://github.com/thesis/mezo-portal/tree/main/solidity#contract-addresses))

---

### Author's checklist

### Reviewer's checklist


jeremyletang (approved):
Looks good to me. Tested locally and works as advertised.

PR #443, opened 2025-04-11
Allow setting bridge validators at genesis

lukasz-zimnoch:
References: https://linear.app/thesis-co/issue/TET-604/mainnet-genesis-file

### Introduction

Here we add a mechanism allowing to pass initial bridge validators at genesis. This allows solving the chicken and-egg problem on mainnet where we need at least one bridge validator to bridge BTC to Mezo but we need to have Mezo BTC to add a bridge validator.

### Changes

We add the new `privilege_assignments` param of the `poa` module genesis allowing to pass bridge validators as follows: 

```
"privilege_assignments": [
    {
        "operator_bech32": "mezovaloper1rrkzqr6nuf9yjkgqhsx22kvruc50ee0wuk4wwu",
        "privilege": "bridge"
    },
    {
        "operator_bech32": "mezovaloper1cgc4v6hcay8tl7qhmmjsc70lh4ewkrxj2wyana",
        "privilege": "bridge"
    }
]
```

Moreover, from now on, we add the first validator as the bridge validator for localnet.

Last but not least, we modify the `mezod export` command so current privilege assignments (so bridge vals as well) are exported from the state with the rest of the state entries.

### Testing

Spin up a fresh localnet. Use `validatorPool:validatorsByPrivilege --privilege-id 1` from `precompile/hardhat` to make sure privileges were added properly. 

Experiment with different values of `poa.privilege_assignments` (empty, multiple, etc.) and ensure all cases work.

Last but not least, you can also check `mezod export`, just remember about setting `--home` to point a localnet node.

---

### Author's checklist

### Reviewer's checklist


x/poa/types/genesis.go
```diff
@@ -74,3 +78,43 @@ func validateGenesisStateValidators(validators []Validator) (err error) {
...
+func validateGenesisStatePrivilegeAssignments(assignments []ValidatorPrivilegeAssignment) error {
+	if len(assignments) == 0 {
+		return nil
+	}
+
+	seen := make(map[string]bool)
+
+	for _, assignment := range assignments {
+		operator := assignment.GetOperatorBech32()
+		privilege := assignment.GetPrivilege()
```
jeremyletang, line 96: Is there a point validating that the privilege string is a valid one, as in it contains e.g: "bridge", not something unsupported? Ultimately it's not a big deal, seeing it comes from the genesis file, you'd expect peer review to find issues early here.

x/poa/types/genesis.go
```diff
@@ -74,3 +78,43 @@ func validateGenesisStateValidators(validators []Validator) (err error) {
...
+//
+// Ensures that:
+// - The operator and privilege are non-empty strings.
+// - There are no duplicated assignments.
+func validateGenesisStatePrivilegeAssignments(assignments []ValidatorPrivilegeAssignment) error {
+	if len(assignments) == 0 {
+		return nil
+	}
+
+	seen := make(map[string]bool)
```
jeremyletang, line 92: nit: I'd use a map[string]struct{} here instead, but it's just for the form / habits, seeing that the map is tiny, and temporary.

jeremyletang (approved):
Looks good to me. Also tested locally and works as advertised:
- with 0 privileges_assignment
- 1
-  many
- using an operator which is not a validator

All works as expected

dimpar:
Same here, also tested with 0, 1 and many validators. Worked fine.

Should we make a note https://linear.app/thesis-co/issue/TET-743/mainnet-chain-rollout to add one so that this task won't disappear from our radar?

dimpar (approved):

x/poa/types/genesis.go
lukasz-zimnoch: Well, actually x/poa doesn't care what the privilege is. It can be any string hence the validation is quite shallow.

PR #444, opened 2025-04-14
Pre-mainnet genesis file

lukasz-zimnoch:
References: https://linear.app/thesis-co/issue/TET-604/mainnet-genesis-file

> [!NOTE]
> I used `mezod keys parse` to convert between EVM and bech32 addresses. Please cross-check using a DIFFERENT tool to make sure addresses encoded in genesis match the ones from the PR description. 

### Introduction

Here we add the genesis file for the pre-mainnet network that will be started in the upcoming days.

### Changes

We created a default genesis for mainnet chain ID `mezo_31612-1` and tagged it with `1.0.0-rc2` which is the planned `mezod` version with pre-mainnet support. Then, we went module by module and changed specific parameters according to the chain needs or left the defaults where applicable. Here is the summary.

#### x/auth

We added 4 accounts that are the initial buckets holding MEZO allocations. Those are pure EOAs. Those accounts are:
- Bucket 1: `0x39D8679fd651689337e6f26aEB0a7B45fb29a5fb`
- Bucket 2: `0x0625C2515A29D09f1Ba4Da5a87BD1cbc67386758`
- Bucket 3: `0xAF1040eC899A9D7355B783aB2727F651a8932996`
- Bucket 4: `0xfA0696b31289D0ec386f032A45B5382230Fda350`

#### x/authz

This module should not have any initial authorizations hence it's section is empty in the genesis file.

#### x/bank

We mint 1B (1,000,000,000) of MEZO and allocate the supply to the aforementioned buckets as follows:
- Bucket 1: 300,000,000 MEZO
- Bucket 2: 200,000,000 MEZO
- Bucket 3: 400,000,000 MEZO
- Bucket 4: 100,000,000 MEZO

#### x/bridge

We configure this module as follows:
- Maximum allowed mappings for the ERC20 bridge is `20`
- We enable the BTC supply assertion (`btc_supply == btc_minted - btc_burnt`)
- We start the `AssetsLocked` sequence tip pointer from `0`
- We set the BTC source token to be mainnet TBTC: `0x18084fbA666a33d37592fA2633fD49a74DD93a88`
- We don't create any ERC20 mappings at genesis
- We set the initial BTC supply to `0`

#### x/crisis

We disabled Cosmos transaction hence this module is a no-op and will be removed in future releases (Cosmos SDK did it as well). We leave default genesis parameters here.

#### x/evm

We configure this module as follows:
- We take the defaults that were used for matsnet
- We allow non-EIP155 (unprotected) transactions from the start to allow deploying SAFE and Passport contracts
- We set the latest version of all 7 precompiles supported by `mezod` (we don't put the test-only Testbed precompile here)
- We allow for max 10 precompile calls in a single transaction
- We leave the chain fee splitter address blank for now as TIgris is not deployed yet

#### x/feemarket

We configure this module as follows:
- We use the default `base_fee_change_denominator`, `elasticity_multiplier`, and `min_gas_multiplier` that were used for matsnet
- We use `24000000 abtc` as the minimum price per gas unit (`min_gas_price`). We computed it with an assumption that a simple transfer taking `21000` of gas should cost `~0.05 USD` assuming `1 BTC = 100,000 USD`. The computed value was rounded up to a human-friendly form to simplify calculations.
- We use `48000000 abtc` (2 * `min_gas_price`) as the initial `base_fee`. This value doesn't really matter as it will be automatically adjusted by the chain. However, we must start with something.

#### x/marketmap and x/oracle

We use the defaults imposed by the Skip Connect library. The address `mezo10d07y265gmmuvt4z0w9aw880jnsr700jdl5p9z` set as `market_authorities` and `admin` is the bech32 address of the Cosmos SDK `x/gov` module. We can leave it as such as any updates here will require a hard fork anyway since this module doesn't have a setter on the EVM layer.

#### x/poa

We configure this module as follows:
- PoA owner is an arbitrary EOA: `0x90cb513eb5b6831a877392Ae6d40b9A9B63155C8`
- Maximum number of validators is `150`
- There are 3 initial validators operated by the initial 3 providers
- The first validator obtains the bridge privilege so it becomes the first bridge validator

#### x/upgrade

We use the defaults for this module.

#### consensus

We configure the consensus with the defaults. We use 10M block gas limit based on the [performance tests](https://github.com/mezo-org/mezod/blob/main/docs/performance-tests.md) we conducted recently. Last but not least, we enable vote extensions on block `1`.

### Testing

You can test this file locally by generating a fresh localnet and pasting localnet validators to the pre-mainnet genesis file. Then, use the pre-mainnet genesis for each localnet validator and run the network.

---

### Author's checklist

### Reviewer's checklist


chain/mainnet/mezo_31612-1/genesis.json
```diff
@@ -1 +1,296 @@
-[]
+{
+  "app_name": "mezod",
+  "app_version": "1.0.0-rc2",
+  "genesis_time": "2025-04-14T08:50:53.767481Z",
```
jeremyletang, line 4: Not that it matters much, but you may want to change this date to one closer to the launch date?

lukasz-zimnoch: We will modify this file and re-genesis for real mainnet so this date will move forward.

jeremyletang: right.

chain/mainnet/mezo_31612-1/genesis.json
```diff
@@ -1 +1,296 @@
...
+    },
+    "authz": {
+      "authorization": []
+    },
+    "bank": {
+      "params": {
+        "send_enabled": [],
+        "default_send_enabled": true
+      },
+      "balances": [
```
dimpar, line 68: Let's double check the buckets and amounts. What we have in the PR description does not align with the order in the `balances` array. Bucket 1 is switched with bucket 2.

`mezo1qcjuy52698gf7xaymfdg00guh3nnse6chm0s2t` -> `0x0625C2515A29D09f1Ba4Da5a87BD1cbc67386758`

`mezo188vx087k295fxdlx7f4wkznmghajnf0maysxpe` -> `0x39D8679fd651689337e6f26aEB0a7B45fb29a5fb`

`mezo14ugypmyfn2whx4dhsw4jwflk2x5fx2vkvmjrzk` -> `0xAF1040eC899A9D7355B783aB2727F651a8932996`

`mezo1lgrfdvcj38gwcwr0qv4ytdfcygc0mg6slsj3yt` -> `0xfA0696b31289D0ec386f032A45B5382230Fda350`

jeremyletang: well, they are not in the same order, but the allocation matches the on in the description, this is what I've  noted when I checked earlier, which match the description:
```
mezo1qcjuy52698gf7xaymfdg00guh3nnse6chm0s2t -> 0x0625c2515a29d09f1ba4da5a87bd1cbc67386758 = 200000000000000000000000000
mezo188vx087k295fxdlx7f4wkznmghajnf0maysxpe -> 0x39d8679fd651689337e6f26aeb0a7b45fb29a5fb = 300000000000000000000000000
mezo14ugypmyfn2whx4dhsw4jwflk2x5fx2vkvmjrzk -> 0xaf1040ec899a9d7355b783ab2727f651a8932996 = 400000000000000000000000000
mezo1lgrfdvcj38gwcwr0qv4ytdfcygc0mg6slsj3yt -> 0xfa0696b31289d0ec386f032a45b5382230fda350 = 100000000000000000000000000
```

dimpar: you're right, all good then

chain/mainnet/mezo_31612-1/genesis.json
```diff
@@ -1 +1,296 @@
...
+      "send_enabled": []
+    },
+    "bridge": {
+      "params": {
+        "max_erc20_tokens_mappings": 20,
+        "btc_supply_assertion_enabled": true
+      },
+      "assets_locked_sequence_tip": "0",
+      "source_btc_token": "0x18084fbA666a33d37592fA2633fD49a74DD93a88",
+      "erc20_tokens_mappings": [],
```
pdyraga, line 122: Are we going to create them manually later? Who needs to perform this transaction?

chain/mainnet/mezo_31612-1/genesis.json
pdyraga, line 68: We may need to separate MEZO minting from genesis. This is a new information from yesterday and we should know it by Wednesday EOD latest. I will try to get the update today.

pdyraga: Nitpicking, but can we get them into the same order? I was confused for a moment just like @dimpar.

chain/mainnet/mezo_31612-1/genesis.json
```diff
@@ -1 +1,296 @@
...
+      "initial_btc_supply": "0"
+    },
+    "crisis": {
+      "constant_fee": {
+        "denom": "abtc",
+        "amount": "1000"
+      }
+    },
+    "evm": {
```
pdyraga, line 131: What do we do about pre-deployment contracts? To save us operational pain later, we could deploy ERC-2470 singleton factory and Multicall3.

We will also need to through the SafeFactory deployment logs and see if the transactions were pre-signed and require a bag of BTC to deploy.

pdyraga: Thinking more about it, the problem is more complicated than it seems. We will have to deploy not only Safe contracts with all their dependencies, but also OrangeKit contracts via an upgrade to v1.0.0-beta.3 per https://github.com/mezo-org/passport/pull/86 to maintain stable addresses. It feels that the best solution is to just return the chain fees once we are done.

chain/mainnet/mezo_31612-1/genesis.json
lukasz-zimnoch: The PoA owner, see [TET-608: Deploy bridgeable ERC20 on Mezo](https://linear.app/thesis-co/issue/TET-608/deploy-bridgeable-erc20-on-mezo)

chain/mainnet/mezo_31612-1/genesis.json
lukasz-zimnoch: This affects the whole pre-mainnet plan if so. Just flagging.

chain/mainnet/mezo_31612-1/genesis.json
lukasz-zimnoch: Yes, pre-deployment of such contracts will be complicated and error prone as we need to work on bare byte code. The refund of chain fees through the splitter address is far easier.

chain/mainnet/mezo_31612-1/genesis.json
dimpar: Can we change the PR description to align with the `genesis` file? Up to you though, really not biggy.

lukasz-zimnoch: The description follows the order from the Linear task that holds the spec. I prefer to follow that order. The order in genesis doesn't really matter for humans.

dimpar (approved):

PR #450, opened 2025-04-23
Improvements to the metrics scrapper

jeremyletang:
Part of: https://linear.app/thesis-co/issue/TET-566/validators-status-page

### Introduction

A few improvement to the metrics scraper

### Changes

- Add retry mechanics on the connection, do not log.Fatal anymore
- Ensure that all metrics are set to default values, even if the node is down.

### Testing

Start a localnode with:
```
make localnode-bin-start
```
Start the metrics scraper using the following config:
```
{
    "poll_rate": "1s",
    "chain_id": "mezo_31611-1",
    "nodes": [
	{
	    "rpc_url": "http://localhost:8545",
	    "moniker": "localnode-1
	}
    ]
}
```

And the following command line:
```
metrics-scraper -config=<PATH_TO_CONFIG>
```

Go to http://localhost:2112/metrics, and find the relevant metrics saying the node is up.

Stop the localnode, and go to the URL again, the node will appear down.
Restart the localnode again, to the URL again, the node will appear up.

### Author's checklist

### Reviewer's checklist


metrics-scraper/monitoring.go
```diff
@@ -163,12 +216,14 @@ func sidecarsVersion(ctx context.Context, client *rpc.Client, moniker, networkID
...
+		// just format nicely the error, and continue execution
+		// so the prometheus variable are still set to defaults
+		err = fmt.Errorf("couldn't call %v for %v: %v", netSidecarsEndpoint, moniker, err)
 	}
 
 	var (
-		ethereumVersion, ethereumIsRunning = "unknow", 0.
-		connectVersion, connectIsRunning   = "unknow", 0.
+		ethereumVersion, ethereumIsRunning = "0.0.0-unknow", 0.
+		connectVersion, connectIsRunning   = "0.0.0-unknow", 0.
```
lukasz-zimnoch, line 226:
```suggestion
		ethereumVersion, ethereumIsRunning = "0.0.0-unknown", 0.
		connectVersion, connectIsRunning   = "0.0.0-unknown", 0.
```

metrics-scraper/monitoring.go
```diff
@@ -69,16 +70,49 @@ func Start(configPath string) {
...
+	wg *sync.WaitGroup,
+	config NodeConfig,
+) (c *rpc.Client, err error) {
+	c, err = rpc.DialContext(ctx, config.RPCURL)
+	if err == nil {
+		// no errors, let's start polling
+		return
+	}
+
+	ticker := time.NewTicker(5 * time.Second) // arbitrarily retry every 5 seconds
```
lukasz-zimnoch, line 86: Let's just use the same `pollRate` from the config as for the regular scraping.

metrics-scraper/monitoring.go
```diff
@@ -69,16 +70,49 @@ func Start(configPath string) {
...
+
 func run(
 	ctx context.Context,
 	wg *sync.WaitGroup,
 	pollRate time.Duration,
 	networkID string,
 	config NodeConfig,
 ) {
-	c, err := rpc.DialContext(ctx, config.RPCURL)
+	c, err := tryConnect(ctx, wg, config)
```
lukasz-zimnoch, line 112: I think there is no need to pass `wg` down to `tryConnect`. It's enough to do `defer wg.Done()` at the beginning of `run` to handle all cases. If we do that. we can also remove the `wg.Done()` call from the `<-ctx.Done()` case within this method.

metrics-scraper/monitoring.go
```diff
@@ -69,16 +70,49 @@ func Start(configPath string) {
 	wg.Wait()
 }
 
+// tryConnect, try to connect to the node forever until it works
+// it'll stop when it acquire a connection or it's asked to stop
+func tryConnect(
+	ctx context.Context,
+	wg *sync.WaitGroup,
+	config NodeConfig,
+) (c *rpc.Client, err error) {
```
lukasz-zimnoch, line 79: Let's get rid of the named return params. They make this fn hard to read and error prone. They are also source of inconsitent behavior: if the context is done before the first iteration of the for loop, the error will be nil but after at least one iteration, it will be non-nil.

Moreover, I think we don't have to return an error from this function but rather `(*rpc.Client, bool)`. If we don't connect and the ctx is done, we return `(nil, false)`. Otherwise, `(c, true)`.

metrics-scraper/monitoring.go
```diff
@@ -91,66 +125,84 @@ func run(
...
 		Timestamp string `json:"timestamp"`
 	}{}
+
 	err := client.CallContext(ctx, &result, ethGetBlockByNumberEndpoint, "latest", false)
 	if err != nil {
 		mezodUpGauge.WithLabelValues(moniker, networkID).Set(0)
+		// set latest block to 0 for error
+		mezoLatestBlockGauge.WithLabelValues(moniker, networkID).Set(0)
+		// set latest timestamp to 0 for error
+		mezoLatestTimestampGauge.WithLabelValues(moniker, networkID).Set(0)
```
lukasz-zimnoch, line 164: I think we are overcomplicating this function. Why not doing the following:
1. We create a variable `err` at the beginning.
2. We create a `defer` that sets all the gauges managed by this fn (latest block and timestamp) to `0` if `err != nil`.
3. Upon an error set `err` and `return`. This way, the defer function will zero the gauges. Of course, we bubble the error up to the caller.

metrics-scraper/monitoring.go
```diff
@@ -91,66 +125,84 @@ func run(
...
 		return fmt.Errorf("couldn't call %v for %v: %v", ethGetBlockByNumberEndpoint, moniker, err)
 	}
 
 	mezodUpGauge.WithLabelValues(moniker, networkID).Set(1)
```
lukasz-zimnoch, line 168: The fact that `mezoUpGauge` is managed inside `latestBlockAndTimestamp` is unfortunate. Let's move it up to `run`. If `pollData` was ok, up = 1. Otherwise, up = 0.

metrics-scraper/monitoring.go
```diff
@@ -69,16 +70,49 @@ func Start(configPath string) {
 	wg.Wait()
 }
 
+// tryConnect, try to connect to the node forever until it works
+// it'll stop when it acquire a connection or it's asked to stop
+func tryConnect(
```
lukasz-zimnoch, line 75: What are the values of the gauges if a node is down and we keep trying to connect?

metrics-scraper/monitoring.go
```diff
@@ -91,66 +125,84 @@ func run(
...
+	} else {
+		mezoLatestTimestampGauge.WithLabelValues(moniker, networkID).Set(float64(ts))
 	}
-	mezoLatestTimestampGauge.WithLabelValues(moniker, networkID).Set(float64(ts))
-
-	return nil
+	return errors.Join(errs...)
 }
 
 func nodeVersion(ctx context.Context, client *rpc.Client, moniker, networkID string) error {
```
lukasz-zimnoch, line 191: I think we can use the same as proposed in https://github.com/mezo-org/mezod/pull/450#discussion_r2057588722 for `nodeVersion`.

metrics-scraper/monitoring.go
```diff
@@ -163,12 +215,14 @@ func sidecarsVersion(ctx context.Context, client *rpc.Client, moniker, networkID
 	result := map[string]net.SidecarInfos{}
```
lukasz-zimnoch, line 215: I think we can use the same as proposed in https://github.com/mezo-org/mezod/pull/450#discussion_r2057588722 for `sidecarsVersion`.

metrics-scraper/monitoring.go
jeremyletang: they'll be empty

metrics-scraper/monitoring.go
jeremyletang: addressed in: d1755782d3115b584b67e6fb763dc3d62386ef81

metrics-scraper/monitoring.go
jeremyletang: addressed in: d1755782d3115b584b67e6fb763dc3d62386ef81

metrics-scraper/monitoring.go
jeremyletang: addressed in: d1755782d3115b584b67e6fb763dc3d62386ef81

metrics-scraper/monitoring.go
jeremyletang: addressed in: d1755782d3115b584b67e6fb763dc3d62386ef81

metrics-scraper/monitoring.go
jeremyletang: addressed in: d1755782d3115b584b67e6fb763dc3d62386ef81

metrics-scraper/monitoring.go
jeremyletang: addressed in: d1755782d3115b584b67e6fb763dc3d62386ef81

metrics-scraper/monitoring.go
jeremyletang: addressed in: d1755782d3115b584b67e6fb763dc3d62386ef81

metrics-scraper/monitoring.go
jeremyletang: addressed in: d1755782d3115b584b67e6fb763dc3d62386ef81

metrics-scraper/monitoring.go
```diff
@@ -159,33 +226,48 @@ func nodeVersion(ctx context.Context, client *rpc.Client, moniker, networkID str
...
+		"chain_id": networkID,
+	})
+
 	result := map[string]net.SidecarInfos{}
-	err := client.CallContext(ctx, &result, netSidecarsEndpoint)
+	err = client.CallContext(ctx, &result, netSidecarsEndpoint)
 	if err != nil {
+		// just format nicely the error, and continue execution
+		// so the prometheus variable are still set to defaults
 		return fmt.Errorf("couldn't call %v for %v: %v", netSidecarsEndpoint, moniker, err)
```
lukasz-zimnoch, line 253: We use named return so let's be consistent:
```suggestion
        err = fmt.Errorf("couldn't call %v for %v: %v", netSidecarsEndpoint, moniker, err)
		return 
```

metrics-scraper/monitoring.go
```diff
@@ -159,33 +226,48 @@ func nodeVersion(ctx context.Context, client *rpc.Client, moniker, networkID str
...
-			connectIsRunning = 1
+			isConnected = 1
 		}
-		connectVersion = connectSidecar.Version
+		connectSidecarGauge.WithLabelValues(moniker, networkID, connectSidecar.Version).Set(isConnected)
 	}
-	connectSidecarGauge.WithLabelValues(moniker, networkID, connectVersion).Set(connectIsRunning)
 
-	return nil
+	return err
```
lukasz-zimnoch, line 272:
```suggestion
	return
```

metrics-scraper/monitoring.go
```diff
@@ -159,33 +226,48 @@ func nodeVersion(ctx context.Context, client *rpc.Client, moniker, networkID str
 	return nil
```
lukasz-zimnoch, line 226:
```suggestion
	return
```

metrics-scraper/monitoring.go
```diff
@@ -69,80 +70,146 @@ func Start(configPath string) {
...
+	mezodVersionGauge.DeletePartialMatch(map[string]string{
+		"moniker":  moniker,
+		"chain_id": networkID,
+	})
+
 	var result string
-	err := client.CallContext(ctx, &result, web3ClientVersionEndpoint)
+	err = client.CallContext(ctx, &result, web3ClientVersionEndpoint)
 	if err != nil {
 		return fmt.Errorf("couldn't call %v for %v: %v", web3ClientVersionEndpoint, moniker, err)
```
lukasz-zimnoch, line 214: Let's be consistent:
```suggestion
        err = fmt.Errorf("couldn't call %v for %v: %v", web3ClientVersionEndpoint, moniker, err)
		return 
```
Same for the error returned for `if len(segments) != 4` below.

metrics-scraper/monitoring.go
```diff
@@ -69,80 +70,146 @@ func Start(configPath string) {
...
 	result := struct {
 		Number    string `json:"number"`
 		Timestamp string `json:"timestamp"`
 	}{}
-	err := client.CallContext(ctx, &result, ethGetBlockByNumberEndpoint, "latest", false)
+
+	err = client.CallContext(ctx, &result, ethGetBlockByNumberEndpoint, "latest", false)
 	if err != nil {
-		mezodUpGauge.WithLabelValues(moniker, networkID).Set(0)
 		return fmt.Errorf("couldn't call %v for %v: %v", ethGetBlockByNumberEndpoint, moniker, err)
```
lukasz-zimnoch, line 175:
```suggestion
        err = fmt.Errorf("couldn't call %v for %v: %v", ethGetBlockByNumberEndpoint, moniker, err)
		return 
```

metrics-scraper/monitoring.go
```diff
@@ -69,80 +70,146 @@ func Start(configPath string) {
...
 	if err != nil {
-		return err
+		errs = append(errs, fmt.Errorf("invalid timestamp: %v", err))
+	} else {
+		mezoLatestTimestampGauge.WithLabelValues(moniker, networkID).Set(float64(ts))
 	}
-	mezoLatestTimestampGauge.WithLabelValues(moniker, networkID).Set(float64(ts))
 
-	return nil
+	return errors.Join(errs...)
```
lukasz-zimnoch, line 193:
```suggestion
	return
```

metrics-scraper/monitoring.go
```diff
@@ -69,80 +70,146 @@ func Start(configPath string) {
...
 	}
-	mezoLatestBlockGauge.WithLabelValues(moniker, networkID).Set(float64(latestBlock))
 
 	ts, err := strconv.ParseUint(strings.TrimPrefix(result.Timestamp, "0x"), 16, 64)
 	if err != nil {
-		return err
+		errs = append(errs, fmt.Errorf("invalid timestamp: %v", err))
+	} else {
+		mezoLatestTimestampGauge.WithLabelValues(moniker, networkID).Set(float64(ts))
 	}
```
lukasz-zimnoch, line 191: I think we should error out on parse. If we cannot parse, something is wrong and we should not display the previous metric value but signal an error with 0:
```suggestion
	latestBlock, err := strconv.ParseUint(strings.TrimPrefix(result.Number, "0x"), 16, 64)
	if err != nil {
		err = fmt.Errorf("invalid latestBlock: %v", err)
        return
	}

	ts, err := strconv.ParseUint(strings.TrimPrefix(result.Timestamp, "0x"), 16, 64)
	if err != nil {
		err = fmt.Errorf("invalid timestamp: %v", err)
        return
	} 

    mezoLatestBlockGauge.WithLabelValues(moniker, networkID).Set(float64(latestBlock))
    mezoLatestTimestampGauge.WithLabelValues(moniker, networkID).Set(float64(ts))
```

metrics-scraper/monitoring.go
jeremyletang: Sure. I don't think they can't be compatible though. Here in those functions I use a named return to show the intent in the defer that we are checking te returned value explicit. Now, actually returning explicitly or not in the function is really just a matter of style. Happy to update the code to your preferred style though.

metrics-scraper/monitoring.go
```diff
@@ -159,33 +226,48 @@ func nodeVersion(ctx context.Context, client *rpc.Client, moniker, networkID str
...
+		"moniker":  moniker,
+		"chain_id": networkID,
+	})
+
 	result := map[string]net.SidecarInfos{}
-	err := client.CallContext(ctx, &result, netSidecarsEndpoint)
+	err = client.CallContext(ctx, &result, netSidecarsEndpoint)
 	if err != nil {
+		// just format nicely the error, and continue execution
+		// so the prometheus variable are still set to defaults
```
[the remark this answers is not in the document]
jeremyletang: addressed in: 94e0597d570fbb1a694bc047ed7b89d8570a1289

metrics-scraper/monitoring.go
jeremyletang: addressed in: 94e0597d570fbb1a694bc047ed7b89d8570a1289

metrics-scraper/monitoring.go
jeremyletang: addressed in: 94e0597d570fbb1a694bc047ed7b89d8570a1289

metrics-scraper/monitoring.go
jeremyletang: addressed in: 94e0597d570fbb1a694bc047ed7b89d8570a1289

metrics-scraper/monitoring.go
jeremyletang: addressed in: 94e0597d570fbb1a694bc047ed7b89d8570a1289

metrics-scraper/monitoring.go
jeremyletang: addressed in: 94e0597d570fbb1a694bc047ed7b89d8570a1289

metrics-scraper/monitoring.go
jeremyletang: addressed in: 94e0597d570fbb1a694bc047ed7b89d8570a1289

lukasz-zimnoch (approved):

lukasz-zimnoch:
LGTM! Will deploy and test on `mezo-staging`.

PR #457, opened 2025-04-25
Generic bridgeable ERC20

dimpar:
Closes: https://linear.app/thesis-co/issue/TET-826/implement-generic-bridgeable-erc20-contract

### Introduction

We need to implement a generic smart contract that can represent ERC20s bridged by the native bridge to Mezo.

There is `TestERC20.sol` contract that inherits from `AbstractERC20Upgradeable.sol`. Once we know all the tokens that should be allowed to bridge, we can replace `TestERC20.sol` with real tokens and inherit from `AbstractERC20Upgradeable.sol`.

Deployment scripts can be also easily modified to reflect the real `ERC20` tokens.

### Testing
As `AbstractERC20Upgradeable.sol` inherits from well known OZ libraries and introduces just a couple of new changes, I've added just a few unit tests for sanity checks. There is no point of testing OZ libraries.

Running unit tests:
`npx hardhat test`

Deployment:
`npx hardhat deploy --network <network>` // hardhat by default
Please see the comments section for the actual `Matsnet` deployment.

---

### Author's checklist

- [x] Provided the appropriate description of the pull request
- [x] Updated relevant unit and integration tests
- [x] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [x] Assigned myself in the `Assignees` field
- [x] Assigned `mezod-developers` in the `Reviewers` field and notified them on Discord

### Reviewer's checklist

- [x] Confirmed all author's checklist items have been addressed
- [x] Considered security implications of the code changes
- [x] Considered performance implications of the code changes
- [x] Tested the changes and summarized covered scenarios and results in a comment

PR #466, opened 2025-05-06
Refactor the bridgeable ERC20 implementation

lukasz-zimnoch:
References: https://linear.app/thesis-co/issue/TET-608/deploy-bridgeable-erc20-on-mezo
Follow-up to: #457

### Introduction

Here we refactor the generic implementation of the bridgeable ERC20 tokens on Mezo.

#### Rename `AbstractERC20Upgradeable` to `mERC20`

The generic implementation was renamed to `mERC20`. This feels reasonable as all ERC20 tokens supported by the native bridge on Mezo will be prefixed with `m*`. 

#### Move `x/bridge/solidity` to `solidity`

Although we decided to place this code inside `x/bridge/solidity` initially, it feels cleaner to move that to a top-level `solidity` directory. If we need to add more contracts in the future (not necessarily ERC20), we have a single place for that.

#### Generic `mERC20` deployment script

We generalized the deployment script so it can be used by multiplr `mERC20` implementations quickly and without duplication.

#### Minor adjustments

Last but not least, we addressed other things, some of them coming from review comments from #457:
- Added `__ERC20Burnable_init` call for consistency
- Made the `mTestERC20.sol` deployment script `hardhat`-only
- Introduced configurable decimals (ERC20 tokens use different values, not only 18)
- Changed the license headers and npm package license to LGPL-3.0 to match the repo license
- Renamed npm package of `solidity` to `@mezo-org/mezod-contracts`
- Aligned the existing HH config in `precompile/hardhat` with `solidity` for consistency (added dotenv)

### Testing

Not applicable. You can just run `npm test` in `solidity`.

---

### Author's checklist

### Reviewer's checklist


solidity/contracts/mERC20.sol
```diff
@@ -1,13 +1,15 @@
...
 import "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20BurnableUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
 
-abstract contract AbstractERC20Upgradeable is ERC20Upgradeable, ERC20BurnableUpgradeable, Ownable2StepUpgradeable {
+abstract contract mERC20 is ERC20Upgradeable, ERC20BurnableUpgradeable, Ownable2StepUpgradeable {
     /// @dev The address of the minter.
     address public minter;
+    /// @dev The number of decimals of the token.
+    uint8 private _decimals;
```
pdyraga, line 12: Why don't we make it public? We would avoid exposing the view function and do just `decimals = _decimals` in the initializer.

solidity/contracts/mERC20.sol
```diff
@@ -32,19 +34,23 @@ abstract contract AbstractERC20Upgradeable is ERC20Upgradeable, ERC20BurnableUpg
...
-        string memory symbol,
-        address initialMinter
+        string memory _name,
+        string memory _symbol,
+        uint8 _decimalsArg,
+        address _minter
     ) public initializer {
-        __ERC20_init(name, symbol);
+        __ERC20_init(_name, _symbol);
+        __ERC20Burnable_init();
```
pdyraga, line 43: Shouldn't we use the unchained version? `ERC20BurnableUpgradeable` inherits from `ERC20Upgradeable` and [based on the docs](https://docs.openzeppelin.com/contracts/4.x/upgradeable#multiple-inheritance) 

> __{ContractName}_init function embeds the linearized calls to all parent initializers.

And it is not something we would do ideally. Right now, `_init` does nothing so we are good. I could argue `_init_unchained` is more future-proof, given it should guarantee `__ERC20_init` will not get called twice. At the same time, using `_init_unchained` is not recommended but... what's the alternative?

Anyway, I am good leaving it as-is, just curious about your opinion.

solidity/contracts/mERC20.sol
```diff
@@ -69,4 +75,8 @@ abstract contract AbstractERC20Upgradeable is ERC20Upgradeable, ERC20BurnableUpg
     function mint(address account, uint256 amount) public onlyMinter {
         _mint(account, amount);
     }
-}
\ No newline at end of file
+
```
pdyraga, line 78: Nitpicking but we could take the opportunity and clean up `setMinter` to just:
```
emit MinterChanged(minter, newMinter);
minter = newMinter;
```

solidity/hardhat.config.ts
```diff
@@ -23,10 +23,15 @@ const config: HardhatUserConfig = {
...
       default: 1,
-      testnet: "0x17F29B073143D8cd97b5bBe492bDEffEC1C5feE5",
-      mainnet: "0x17F29B073143D8cd97b5bBe492bDEffEC1C5feE5",
+      testnet: "0x6e80164ea60673d64d5d6228beb684a1274bb017", // testertesting.eth
+      mainnet: "0x98D8899c3030741925BE630C710A98B57F397C7a"  // mezo multisig
+    },
+    minter: {
+      default: 2,
+      testnet: "0x17F29B073143D8cd97b5bBe492bDEffEC1C5feE5", // x/bridge module account
+      mainnet: "0x17F29B073143D8cd97b5bBe492bDEffEC1C5feE5", // x/bridge module account`
```
pdyraga, line 34: What's the best way to verify this address?

solidity/test/mTestERC20.test.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    it("should set the correct minter", async function () {
+      const { mTestERC20, minter } = await loadFixture(fixture)
+      expect(await mTestERC20.minter()).to.equal(minter)
+    })
+
+    it("should set the correct owner", async function () {
+      const { mTestERC20, owner } = await loadFixture(fixture)
+      expect(await mTestERC20.owner()).to.equal(owner.address)
+    })
```
pdyraga, line 45: Should we add a test for setting decimals?

pdyraga (commented):
Some minor questions and comments and I am good to merge.

solidity/contracts/mERC20.sol
lukasz-zimnoch: Because we need to explicitly overwrite the function from the upstream contract.

solidity/contracts/mERC20.sol
lukasz-zimnoch: As you said, the general OZ recommendation is not using `unchained`. I'm just following it 🤷

solidity/hardhat.config.ts
lukasz-zimnoch: See https://github.com/mezo-org/mezod/pull/457#issuecomment-2838323949. However, we recently removed `mezod debug` CLI so use `mezod keys parse` instead.

solidity/test/mTestERC20.test.ts
lukasz-zimnoch: We can 182fd47f8fb356e20071550a9f4906c50cee38b9

solidity/contracts/mERC20.sol
lukasz-zimnoch: The zero-address check doesn't hurt. Also, I already did the testnet deployment in https://github.com/mezo-org/mezod/pull/467 and I'd prefer to not waste time to re-deploy everything and modify the mappings if not really needed.

lukasz-zimnoch:
@pdyraga all comments addressed!

solidity/contracts/mERC20.sol
pdyraga: I meant the extra `address oldMinter = ` and leave the zero address check. This extra field is not needed; we can just emit the event before. But sure, it's a minor-minor and if we already deployed them, there is no need to waste time.

solidity/contracts/mERC20.sol
pdyraga: Ah, right.

pdyraga (approved):

PR #475, opened 2025-05-20
Fix behaviour of the ERC20.approve(0)

jeremyletang:
close: https://linear.app/thesis-co/issue/TET-999/approve-in-erc20-reverts-when-making-a-zero-call-if-the-current

### Introduction

The ERC20 std implementation (e.g: openzeppelin) allow a user to approve 0 with success. This is to allow such expected pattern where a user would first call approve(0) then approve(actualAmount) on the contract. Until now the Mezo precompile would actually return an error an revert if the allowance was already 0.

This PR fixes this behaviour, and basically makes a call to approve(0) a no op effectively.

### Changes

We don't return an error anymore if the allowance of the user is already 0.

### Testing

2 system tests have been added to cover:
- allowing 0 for the very first time
- allowing for an amount initially, then allowing 0, then allowing 0 again.

### Author's checklist

### Reviewer's checklist

### DIFF
--- a/precompile/erc20/approve.go
+++ b/precompile/erc20/approve.go
@@ -135,7 +135,7 @@ func handleAuthorization(
 
 	if authorization == nil {
 		if amount.Sign() == 0 {
-			err = fmt.Errorf("no existing approvals, cannot approve 0")
+			err = nil // this is not an error to comply with the ERC20 std behaviour.
 		} else {
 			err = createAuthorization(context.SdkCtx(), denom, spender, granter, amount, authzkeeper)
 		}
--- a/precompile/erc20/approve.go
+++ b/precompile/erc20/approve.go
@@ -1,6 +1,7 @@
 package erc20
 
 import (
+	"errors"
 	"fmt"
 	"math/big"
 	"time"
@@ -90,8 +91,13 @@ func (am *ApproveMethod) Run(
 	if !ok {
 		return nil, fmt.Errorf("invalid amount: %v", inputs[1])
 	}
-	if amount == nil || amount.Sign() < 0 {
-		amount = big.NewInt(0)
+
+	if amount == nil {
+		return nil, errors.New("amount is required")
+	}
+
+	if amount.Sign() < 0 {
+		return nil, errors.New("amount cannot be negative")
 	}
 
 	granter := context.MsgSender()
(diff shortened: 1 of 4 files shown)


tests/system/contracts/BTCTransfers.sol
```diff
@@ -29,6 +29,12 @@ contract BTCTransfers {
     address private constant testbedPrecompile = 0x7b7c100000000000000000000000000000000000;
     uint256 public balanceTracker = 1;
 
+
+    function approveZeroDoesntReverts(address recipient) external payable {
+        bool ok = IBTC(precompile).approve(recipient, 0);
+	require(ok);
+    }
```
lukasz-zimnoch: Do we need it?

tests/system/test/BTCTransfers.test.ts
```diff
@@ -28,6 +28,65 @@ describe("BTCTransfers", function () {
...
+    let tx2: any;
+    let tx3: any;
+
+    before(async function () {
+      await fixture();
+
+      // here we try to first do an approve,
+      // then reset it with 0,
+      // then approve again with 0, before this would
+      // panic
```
lukasz-zimnoch, line 67:
```suggestion
      // Here we try to first do an approve, then reset it with 0, then approve again with 0.
      // All transactions should pass.
```

precompile/erc20/approve.go
```diff
@@ -135,7 +135,7 @@ func handleAuthorization(
 
 	if authorization == nil {
 		if amount.Sign() == 0 {
-			err = fmt.Errorf("no existing approvals, cannot approve 0")
+			err = nil // this is not an error to comply with the ERC20 std behavior.
```
lukasz-zimnoch, line 138: Note that `amoun.Sign() == 0` also covers the case where someone calls it with a negative value. We default to zero in that case in `ApproveMethod.Run`. Would be good to explicitly fail for negatives.

lukasz-zimnoch: Same for `PermitMethod` that uses `handleAuthorization`.

jeremyletang: I do not believe so as per the code?
```
   40// Sign returns:
   41//   - -1 if x < 0;
   42//   - 0 if x == 0;
   43//   - +1 if x > 0.
   44func (x *Int) Sign() int {
```

jeremyletang: addressed in: 03396387d80d58ae442524ccc8dced82489afae2

lukasz-zimnoch (dismissed):
Works as advertised and system tests pass!

lukasz-zimnoch (approved):

PR #481, opened 2025-05-22
add panics when converting AsTransaction as the transaction should always be valid

jeremyletang:
close: https://linear.app/thesis-co/issue/TET-1033/missing-nil-checks-for-astransaction-may-cause-node-crash

### Introduction

Add panics on codepath were ethereum transaction could be incomplete. This should never happen, but we panic in case the transaction are nil in order to leave time for node operator to investigate the issue.

### Changes

Adds nil checks and panics.

### Testing

N/A

### Author's checklist

### Reviewer's checklist


dimpar (approved):
Leaving a final check for Łukasz.

rpc/backend/utils.go
lukasz-zimnoch, line 1: There are multiple usages of `AsTransaction`. We may want to panic from the inside this function instead of returning `nil`.

PR #506, opened 2025-07-01
`AssetsUnlocked` monitoring loop: Ethereum sidecar side

tomaszslabon:
References: https://linear.app/thesis-co/issue/TET-1182/ethereum-sidecar-monitoring-loop

### Introduction

This PR enables gathering information on `AssetsUnlocked` events from validator nodes to Ethereum sidecar.
The Ethereum sidecar must learn about the state of `AssetsUnlocked` on the Mezo chain.
As a result of this PR, the Ethereum sidecar has a queue of `AssetsUnlocked`  events that should be attested in the `MezoBridge` contract on Ethereum.

### Changes

- Added communication with an endpoint enabling sending requests for `AsssetsUnlocked` events and sequence tip
- Added fetching of information from Mezo chain: recent (~1 month) `AssetsUnlocked` events
- Added fetching of information from `MezoBridge` on Ethereum: `AssetsUnlockConfirmed` events
- Implemented the actual observation routine - storing `AssetsUnlocked` events in the attestation queue

### Testing

Covered with unit tests.

---

### Author's checklist

- [X] Updated relevant unit and integration tests

### Reviewer's checklist


lukasz-zimnoch:
@tomaszslabon please note there is an `Unverified` commit here.

ethereum/sidecar/bridge_out_server.go
```diff
@@ -0,0 +1,101 @@
...
+func (bos *BridgeOutServer) startGRPCServer(
+	ctx context.Context,
+	address string,
+) error {
+	listener, err := net.Listen("tcp", address)
+	if err != nil {
+		return fmt.Errorf("failed to listen: [%w]", err)
+	}
+
+	pb.RegisterEthereumSidecarBridgeOutServiceServer(bos.grpcServer, bos)
```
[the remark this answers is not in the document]
lukasz-zimnoch: Why not using JSON-RPC exposed by `mezod`? `AssetsUnlocked` events will be saved in the chain state (see [TET-1180: AssetsBridge precompile: The `bridgeOut` method](https://linear.app/thesis-co/issue/TET-1180/assetsbridge-precompile-the-bridgeout-method)). I think we can expose a new method under `mezo_` namespace that will return those events (e.g. `mezo_assetsUnlocked`).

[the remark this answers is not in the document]
jeremyletang: Why not just pushed the assetUnlocks directly to the sidercar instead of having it polling for them? I realise we do the inverse mecanism from for the AssetLocks, yet, having both the sidecar and the node have some kind of "circular" dependencies, sounds a bit strange to me.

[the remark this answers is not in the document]
lukasz-zimnoch: I suggested JSON-RPC because I had other use cases in mind (like dApps or indexers wanting to know about assets unlocks). However, we will emit an event from the `AssetsBridge.bridgeOut` function hence those cases should be satisfied anyway. That said, I think your arguments for using gRPC are good enough to go that way.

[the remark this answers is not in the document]
jeremyletang: yes indeed, the sidecar would save first the assetUnlocked, in it's state upon reception, then return a success / error state. Then asynchronously the sidecar will process the new entry.

I'm not picky either way, I just feel like having two service depending of each other explicitly seems a bit messy.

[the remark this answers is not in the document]
lukasz-zimnoch: To add more arguments for the "pull" approach so you can feel better 😅:
- Cleaner boundaries. The Mezo chain just burns the token and stores `AssetsUnlocked` in the state. The sidecar takes care of the rest.
- Less logic in the `mezod` chain client. We already have a lot happening in this process - the less we put here, the better.
- If we need to upgrade this logic, having it in the sidecar makes the upgrade easier
- This approach feels simpler logic-wise. Sidecar just cross-checks what `AssetsUnlocked` are on Mezo and what was already processed on Ethereum then act appropriately.

ethereum/sidecar/bridge_out_client.go
```diff
@@ -0,0 +1,181 @@
...
+				"ethereum sidecar bridge-out connection test failed; possible " +
+					"problem with sidecar configuration or connectivity",
+			)
+		} else {
+			logger.Info(
+				"ethereum bridge-out sidecar connection test completed " +
+					"successfully",
+			)
+		}
+	}()
```
jeremyletang, line 80: Is that goroutine necessary, seeing that it only log the result, we could simply log error any following connection issue instead.

ethereum/sidecar/bridge_out_client.go
```diff
@@ -0,0 +1,181 @@
...
+func NewBridgeOutGrpcClient(
+	logger log.Logger,
+	serverAddress string,
+	requestTimeout time.Duration,
+	registry types.InterfaceRegistry,
+) (*BridgeOutGrpcClient, error) {
+	connection, err := grpc.NewClient(
+		serverAddress,
+		// TODO: Consider using TLS protocol so that the Mezo node and Ethereum
+		//       sidecar can be run on separate servers.
```
jeremyletang, line 37: Is that necessary? I would expect those two to run on private archictecture without accessibility from the outside. This should be the responsibility of the node provider to secure this their own way I would say.

ethereum/sidecar/bridge_out_client.go
```diff
@@ -0,0 +1,181 @@
...
+// GetAssetsUnlockedEvents gets the AssetsUnlocked events from the Mezo chain.
+// The requested range of events is inclusive on the lower side and exclusive
+// on the upper side.
+func (c *BridgeOutGrpcClient) GetAssetsUnlockedEvents(
+	ctx context.Context,
+	sequenceStart sdkmath.Int,
+	sequenceEnd sdkmath.Int,
+) ([]bridgetypes.AssetsUnlockedEvent, error) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
```
jeremyletang, line 94: I may be missing something but why do we need a mutex here?

ethereum/sidecar/bridge_out_client.go
```diff
@@ -0,0 +1,181 @@
...
+		SequenceEnd:   sequenceEnd,
+	}
+
+	response, err := queryClient.AssetsUnlockedEvents(ctxWithTimeout, request)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get AssetsUnlocked events: [%w]", err)
+	}
+
+	events := make([]bridgetypes.AssetsUnlockedEvent, len(response.Events))
+	copy(events, response.Events)
```
jeremyletang, line 112: Why the copy here, and not return directly the slice?

ethereum/sidecar/bridge_out_client.go
```diff
@@ -0,0 +1,181 @@
...
+
+// GetAssetsUnlockedSequenceTip gets the assets unlocked sequence tip from the
+// Mezo chain. The returned sequence tip is equal to the number of AssetsUnlocked
+// events made so far. It is also equal to the value of the unlock sequence
+// in the newest AssetsUnlocked event.
+func (c *BridgeOutGrpcClient) GetAssetsUnlockedSequenceTip(
+	ctx context.Context,
+) (sdkmath.Int, error) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
```
jeremyletang, line 130: Same question here for the mutex? Is it because of hte queryClient? I'd expect this to be usable from multiple routines at the same time, and if it's not the case, the locking should be done inside the querServer implementation.

ethereum/sidecar/cli/flags.go
```diff
@@ -61,5 +66,19 @@ func NewFlagSetEthereumSidecar(
 		"Requests per minute for an Ethereum RPC provider",
 	)
 
+	fs.String(
+		FlagBridgeOutServerAddress,
+		defaultBridgeOutServerAddress,
+		"Address of the gRPC bridge-out server. The bridge-out server is "+
+			"run by validators as part of mezod and responds with "+
+			"AssetsUnlock events",
```
jeremyletang, line 74: Is the concatenation here to make a multiline string? I'd use ` strings instead then for readability.

ethereum/sidecar/server.go
```diff
@@ -45,6 +46,19 @@ var (
 	// enough.
 	cachedEventsLimit = 500000
 
+	// bridgeOutLookBackPeriod is the look-back period used when fetching
+	// AssetsUnlocked events from the Mezo chain. It defines how far back we
+	// look when searching for unattested events.
+	bridgeOutLookBackPeriod = 30 * 24 * time.Hour // ~1 month
```
jeremyletang, line 52: I realised i's a bit too late seeing that you have added the timestamps field in the AssetsUnlock events. But why use time instead of blockHeight for this? For some reason, the later seems more relevant to me when speaking of a blockchain. It's not a big deal though.

ethereum/sidecar/server.go
```diff
@@ -53,12 +67,35 @@ var (
...
+	// exclusive on the upper side.
+	GetAssetsUnlockedEvents(
+		ctx context.Context,
+		sequenceStart sdkmath.Int,
+		sequenceEnd sdkmath.Int,
+	) ([]bridgetypes.AssetsUnlockedEvent, error)
+}
+
+// TimeFunc is a function that returns the current time.
+type TimeFunc func() time.Time
```
jeremyletang, line 89: I'm assuming this is done for testing purpose. I've usually been used to instead create a time package (e.g: mezotime) in the project I work on, where I'd define something like:
```
var Now = time.Now
```

Then in the code use mezotime.Now() when need the time. during normal execution of the software, nothing to do, but when testing, before your test, you can just easily replace the function with a mocked one. Not saying it's a better solution or anything though.

ethereum/sidecar/server.go
```diff
@@ -53,12 +67,35 @@ var (
...
+
+// Server enables exchange of bridging-related information between the Mezo
+// and Ethereum chains. It used for both bridging-in of assets from Ethereum
+// to Mezo and bridging-out the other way around. It is intended to be run as
```
jeremyletang, line 93: nit: That could probably be reworded a little bit.

ethereum/sidecar/cli/flags.go
```diff
@@ -12,6 +13,8 @@ const (
 	FlagServerNetwork             = "ethereum-sidecar.server.network"
 	FlagServerBatchSize           = "ethereum-sidecar.server.batch-size"
 	FlagServerRequestsPerMinute   = "ethereum-sidecar.server.requests-per-minute"
+	FlagBridgeOutServerAddress    = "ethereum-sidecar.bridge-out.server-address"
+	FlagBridgeOutRequestTimeout   = "ethereum-sidecar.bridge-out.request-timeout"
```
lukasz-zimnoch, line 17: I'm not sure if we should make this timeout configurable. I don't see any reason an operator would want to change that to be honest.

ethereum/sidecar/cli/flags.go
```diff
@@ -12,6 +13,8 @@ const (
 	FlagServerNetwork             = "ethereum-sidecar.server.network"
 	FlagServerBatchSize           = "ethereum-sidecar.server.batch-size"
 	FlagServerRequestsPerMinute   = "ethereum-sidecar.server.requests-per-minute"
+	FlagBridgeOutServerAddress    = "ethereum-sidecar.bridge-out.server-address"
```
lukasz-zimnoch, line 16: "Bridge out server" sounds very cryptic. Also, we now have too much clients and servers within the `sidecar` package. What do you think about naming this flag `ethereum-sidecar.assets-unlocked-endpoint`. This way we will be more specific. This will be also consistent with Connect sidecar where they use `market-map-endpoint` which will be easier to explain to operators.

If we agree about that, let's change the entire nomenclature accordingly, so, for example:
- `bridge_out_client.go` becomes just `assets_unlocked.go`
- `BridgeOutGrpcClient` becomes `AssetsUnlockedEndpoint`
- `bridgeOutLookBackPeriod` becomes `AssetsUnlockedLookbackPeriod`

ethereum/sidecar/server.go
```diff
@@ -45,6 +46,19 @@ var (
...
 	cachedEventsLimit = 500000
 
+	// bridgeOutLookBackPeriod is the look-back period used when fetching
+	// AssetsUnlocked events from the Mezo chain. It defines how far back we
+	// look when searching for unattested events.
+	bridgeOutLookBackPeriod = 30 * 24 * time.Hour // ~1 month
+
+	// assetsUnlockedBatchSize is the number of AssetsUnlocked events we can
+	// fetch from the Mezo blockchain in one gRPC call.
+	assetsUnlockedBatchSize = 100
```
lukasz-zimnoch, line 56: Why 100? IIRC we introduced 10k as maximum batch so at least 1000 sounds reasonable.

ethereum/sidecar/server.go
```diff
@@ -45,6 +46,19 @@ var (
...
+
+	// assetsUnlockedFetchingPeriod is the time period defining how often new
+	// AssetsUnlocked events should be fetched.
+	assetsUnlockedFetchingPeriod = 5 * time.Minute
```
lukasz-zimnoch, line 60: I would use 1 minute at most. This communication is cheap and Mezo has instant block finality. Why adding an artificial bottleneck in that case?

lukasz-zimnoch (commented):
First high-level pass. Haven't dug into the logic yet.

ethereum/sidecar/server.go
```diff
@@ -428,6 +540,342 @@ func (s *Server) startGRPCServer(
...
+
+	// Walk backwards from the current sequence tip in windows of at most
+	// `assetsUnlockedBatchSize`.
+outer:
+	for sequenceTip.IsPositive() {
+		// Sequence end is exclusive. We must add `1`.
+		seqEnd := sequenceTip.AddRaw(1)
+
+		// Sequence start is inclusive. It cannot be lower than `1`.
+		seqStart := sequenceTip.SubRaw(int64(s.assetsUnlockedBatchSize - 1))
```
lukasz-zimnoch, line 653: Would it be simpler to do:
```suggestion
		seqStart := seqEnd.SubRaw(int64(s.assetsUnlockedBatchSize))
```

ethereum/sidecar/bridge_out_client.go
tomaszslabon: Removed the mutex in d2f84005e08737970de7d06e8d9b1d5f5e626ab3.
This code should be thread-safe even if functions were executed in parallel.

ethereum/sidecar/bridge_out_client.go
tomaszslabon: Removed the mutex in d2f84005e08737970de7d06e8d9b1d5f5e626ab3.
It's not needed, indeed.

ethereum/sidecar/bridge_out_client.go
tomaszslabon: Agree. Removed the TODO in fd27a6f7e9f5c2055268eb1aba6cae8f4bcb91c3.

ethereum/sidecar/server.go
```diff
@@ -428,6 +540,342 @@ func (s *Server) startGRPCServer(
...
+		if int64(len(events)) != expectedLength {
+			return nil, fmt.Errorf(
+				"fetched unexpected number of AssetsUnlocked events for "+
+					"range [%s, %s); expected %d, got %d",
+				seqStart.String(),
+				seqEnd.String(),
+				expectedLength,
+				len(events),
+			)
+		}
```
lukasz-zimnoch, line 684: Just checking my understanfing. We know that `GetAssetsUnlockedEvents` guarantees we receive a strictly increasing sequence of events and that those events are in the requested boundaries. This check makes sure that `sequence_start == first_event` and `last_event == sequence_end - 1` . This combined with the sort at the end of this function guarantees that `fetchRecentAssetsUnlockedEvents` always return strictly increasing sequences, after the windows are glued together. Is that correct?

If so, I'm wondering if we really need this check. After all, we are asking for sequence ranges that we know for sure have corresponding events (because we are moving backwards from sequence tip). If there are gaps here, this means a critical issue on the mezod side. I don't say we should protect against it but this is probably a wrong place to do so. If an expected event for the given sequence does not exist, we should error out in the mezod query server. I believe we already do it but let's double-check that all I'm saying here makes sense.

ethereum/sidecar/server.go
```diff
@@ -121,25 +176,43 @@ func RunServer(
...
 	}
 
+	bridgeOutClient, err := NewBridgeOutGrpcClient(
+		logger,
+		bridgeOutServerAddress,
+		bridgeOutRequestTimeout,
+		registry,
+	)
+	if err != nil {
+		panic(fmt.Sprintf("failed to create bridge-out client: %v", err))
```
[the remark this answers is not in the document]
jeremyletang: interesting.

ethereum/sidecar/bridge_out_client.go
tomaszslabon: Removed in 0fabe4b90efb7574367b632b0ed8f28333bd1b04.
If there is a connection problem, we will quickly learn about it anyway.

ethereum/sidecar/cli/flags.go
tomaszslabon: I wrote it that way to avoid creating a vary long line of code (say `80` chars in length).

ethereum/sidecar/server.go
```diff
@@ -428,6 +537,342 @@ func (s *Server) startGRPCServer(
...
+	finalizedBlock, err := s.chain.FinalizedBlock(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the finalized block: [%w]", err)
+	}
+	endBlock := finalizedBlock.Uint64()
+
+	// Search for `AssetsUnlockConfirmed` events for the given unlock sequences.
+	// Limit the search to finalized blocks.
+	confirmedEvents, err := s.bridgeContract.PastAssetsUnlockConfirmedEvents(
+		0, // TODO: Would setting startBlock to a lower value improve performance?
```
tomaszslabon, line 739: When searching for recent `AssetsUnlocked` events on Mezo we look back `~1 month`.
We could do the same here - limit the search for `AssetsUnlockConfirmed` events on Ethereum to `~1 month`, since `AssetsUnlockConfirmed` for the given entry can only happen after such events is created on Mezo.

The problem, however, is we need to have a block number, rather than a timestamp. We can estimate which block on Ethereum was 1 month ago, but we could in theory miss some `AssetsUnlockConfirmed` events if the blocks on Ethereum were particularly short recently. Therefore, we would need some margin, e.g. look `1.5 months` back.

ethereum/sidecar/server.go
```diff
@@ -152,12 +222,46 @@ func RunServer(
 		server.logger.Info("gRPC server routine stopped")
 	}()
 
+	go func() {
+		// TODO: Only a subset of validators should be attesting AssetsUnlocked
+		//       events. Decide whether we should run this goroutine by calling
+		//       MezoBridge.bridgeValidators() with our Ethereum address.
```
tomaszslabon, line 228: Could it be as simple as adding a keystore to the base chain, similar to what we did in [keep-core](https://github.com/threshold-network/keep-core/blob/406cb8954005484665422c9693dc780a7ca2d425/pkg/chain/ethereum/ethereum.go#L37)?
The keystore has an `Address` field telling us what is the address of our account, right @lukasz-zimnoch?
We could then call `MezoBridge.bridgeValidators()` with that address to see if we should be attesting.

lukasz-zimnoch: Probably but this is a seprate task: [TET-1383: Ethereum sidecar: Don't run monitoring and attestation loops for non-bridge validators](https://linear.app/thesis-co/issue/TET-1383/ethereum-sidecar-dont-run-monitoring-and-attestation-loops-for-non). Let's not focus on it here.

ethereum/contracts.go
```diff
@@ -22,4 +22,50 @@ type BridgeContract interface {
...
+	TxIndex uint `json:"transactionIndex" rlp:"-"`
+	// hash of the block in which the transaction was included
+	BlockHash common.Hash `json:"blockHash" rlp:"-"`
+	// index of the log in the block
+	Index uint `json:"logIndex" rlp:"-"`
+
+	// The Removed field is true if this log was reverted due to a chain reorganization.
+	// You must pay attention to this field if you receive logs through a filter query.
+	Removed bool `json:"removed" rlp:"-"`
 }
```
lukasz-zimnoch, line 71: Do we really need to map the entire `Raw Log` field? We probably need the content of the AssetsUnlocked and maybe some block/time info, not the entire struct.

ethereum/contracts.go
```diff
@@ -22,4 +22,50 @@ type BridgeContract interface {
 		recipient []common.Address,
 		token []common.Address,
 	) (AssetsLockedIterator, error)
+
+	PastAssetsUnlockConfirmedEvents(
```
lukasz-zimnoch, line 26: Shouldn't this just be:
```suggestion
	PastAssetsUnlockEvents(
```
How this method is supposed to ensure the events are confirmed? Isn't it about taking the right block range (i.e. finalized) on Ethereum?

ethereum/contracts.go
```diff
@@ -22,4 +22,50 @@ type BridgeContract interface {
...
+		endBlock *uint64,
+		unlockSequenceNumberFilter []*big.Int,
+		recipientFilter [][]byte,
+		tokenFilter []common.Address,
+	) ([]*MezoBridgeAssetsUnlockConfirmed, error)
+
+	ConfirmedUnlocks(arg0 *big.Int) (bool, error)
+}
+
+type MezoBridgeAssetsUnlockConfirmed struct { // TODO: Remove once bindings for `MezoBridge` are re-generated and contain features related to attestation.
```
lukasz-zimnoch, line 37: I think the `Confirmed` suffix should not be part of the struct name. This is just a bag of data, the context is added one level-above.
```suggestion
type MezoBridgeAssetsUnlocked struct { // TODO: Remove once bindings for `MezoBridge` are re-generated and contain features related to attestation.
```

ethereum/sidecar/server.go
lukasz-zimnoch: A wide block span is always problematic for Ethereum RPC as providers typically impose 10k block limits and we need to use partial fetch. However, in this particular case, shouldn't we rely only on the sequences filter? Let's also split that to batches just in case and fetch per e.g. 1000 sequences (many providers limit responses to 10k logs). Then, in the next steps, we consider all events with block LTE finalizedBlock as confirmed and those with block GT finalizedBlock as unconfirmed. WDYT?

ethereum/sidecar/server.go
```diff
@@ -428,6 +537,342 @@ func (s *Server) startGRPCServer(
...
+	if err != nil {
+		return nil, fmt.Errorf(
+			"failed to fetch past AssetsUnlockConfirmed events: [%w]",
+			err,
+		)
+	}
+
+	// TODO: We might need to split getting PastAssetsUnlockConfirmedEvents into batches.
+
+	attestedUnlockSequences := make(map[string]bool)
```
lukasz-zimnoch, line 754: Let's use the word "confirmed". Attestation is a short-living process during which validators provide their attestations. Even gathering enough attestations is marked as confirmed.

ethereum/sidecar/server.go
```diff
@@ -428,6 +537,342 @@ func (s *Server) startGRPCServer(
...
+		}
+
+		if isConfirmed {
+			attestedUnlockSequences[confirmedUnlockSequence.String()] = true
+		}
+	}
+
+	// If the event's unlock sequence is not among the attested sequences
+	// consider the event unattested.
+	unattestedEvents := []bridgetypes.AssetsUnlockedEvent{}
```
lukasz-zimnoch, line 776: Let's use "unconfirmed"

ethereum/sidecar/server.go
```diff
@@ -428,6 +537,342 @@ func (s *Server) startGRPCServer(
...
+	})
+
+	return recentEvents, nil
+}
+
+// findUnattestedAssetsUnlockedEvents finds unattested `AssetsUnlocked` events
+// from the passed input events. An event is considered attested if there
+// was an `AssetsUnlockConfirmed` event emitted for it on Ethereum and its
+// unlock sequence is among the confirmed unlocks.
+func (s *Server) findUnattestedAssetsUnlockedEvents(
```
lukasz-zimnoch, line 713: See previous comments:
```suggestion
func (s *Server) findUnconfirmedAssetsUnlockedEvents(
```
Let's update the docstring accordingly.

ethereum/sidecar/server.go
```diff
@@ -428,6 +537,342 @@ func (s *Server) startGRPCServer(
...
+
+	s.attestationMutex.Lock()
+	s.attestationQueue = unattestedEvents
+	s.attestationMutex.Unlock()
+
+	// Save the unlock sequence of the last event as the starting point for
+	// further event fetching.
+	s.lastAssetsUnlockedSequence = sdkmath.ZeroInt()
+	if len(recentEvents) > 0 {
+		s.lastAssetsUnlockedSequence = recentEvents[len(recentEvents)-1].UnlockSequence
```
lukasz-zimnoch, line 582: Double-checking my understanding: We are not caring about unconfirmed events here becauses we pushed them to the attestation queue. We do not re-check them later as this is the responsibility of the attestation loop. Is that correct?

ethereum/sidecar/server.go
```diff
@@ -428,6 +537,342 @@ func (s *Server) startGRPCServer(
...
+				"the last processed unlock sequence (%s vs %s)",
+			sequenceTip.String(),
+			s.lastAssetsUnlockedSequence.String(),
+		)
+	}
+
+	// If the unlock sequence tip has not advanced on Mezo return early.
+	if sequenceTip.Equal(s.lastAssetsUnlockedSequence) {
+		s.logger.Info(
+			"No new AssetsUnlocked events to fetch from Mezo",
```
lukasz-zimnoch, line 814: Nitpick:
```suggestion
			"no new AssetsUnlocked events to fetch from Mezo",
```

ethereum/sidecar/server.go
```diff
@@ -428,6 +540,342 @@ func (s *Server) startGRPCServer(
...
+	}
+
+	// The unlock sequence tip on Mezo is greater than the unlock sequence from
+	// the last fetched event. We need to fetch events from Mezo.
+	seqStart := s.lastAssetsUnlockedSequence.AddRaw(1)
+	seqEnd := sequenceTip.AddRaw(1)
+
+	// TODO: Do we need to split the request into batches?
+	//       Fetching new events is run frequently - there shouldn't be
+	//       many events to fetch.
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah I agree. More than 10k events per minute is rather unrealistic. Even if that happen, the fix would be rather simple: increase the limit and check more often. Let's drop a clear CAPSLOCK comment about this limitation here.

ethereum/sidecar/server.go
```diff
@@ -428,6 +537,342 @@ func (s *Server) startGRPCServer(
...
+	if int64(len(events)) != expectedLength {
+		return fmt.Errorf(
+			"fetched unexpected number of AssetsUnlocked events for "+
+				"range [%s, %s); expected %d, got %d",
+			seqStart.String(),
+			seqEnd.String(),
+			expectedLength,
+			len(events),
+		)
+	}
```
lukasz-zimnoch, line 852: Same comment as in https://github.com/mezo-org/mezod/pull/506#discussion_r2257570132

ethereum/sidecar/server.go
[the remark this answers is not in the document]
tomaszslabon: Added an explanatory comment in 183331f992983ffcb2930ca512ac875351b7e80c.

ethereum/sidecar/server.go
tomaszslabon: Done in a49106d2e537f62103c0da7fa612902c6426af29.

ethereum/sidecar/server.go
tomaszslabon: Done in 0d2144781d4d683df2c171302b0c640049659133.

ethereum/sidecar/server.go
tomaszslabon: Done in 0d2144781d4d683df2c171302b0c640049659133.

ethereum/sidecar/server.go
tomaszslabon: Yes.
In my opinion, the only job of the monitoring loop is to push unconfirmed events to the attestation queue.
Only the attestation loop should car about removing them once they are confirmed.

ethereum/sidecar/cli/flags.go
tomaszslabon: Done in 389206bf39397eedfe2187cf18c45c15eeca13fb, 695e7e99bbe88871b97a2e31a7858226ec5185fa, ee74efe8e72f4d7c35b1be9967356a481297ec03.

ethereum/contracts.go
tomaszslabon: This whole structure is not really needed and will be removed once we generate bindings for `MezoBridgeV2`.
I simply compiled MezoBridgeV2 locally and copied the whole structure from the artifacts to make everything compile. I suggest we keep the copied code here unchanged and remove it once the bindings are generated.
(There is a TODO for this).

ethereum/contracts.go
tomaszslabon: We cannot name this method `PastAssetsUnlockEvents`. We need to keep it as `PastAssetsUnlockConfirmedEvents` as this is the name that will be generated when the package for MezoBridgeV2 is released. Notice that all the functions in this interface are simply functions in the generated bindings.

> How this method is supposed to ensure the events are confirmed?
This method does not guarantee anything on its own, it's a pure connection to the MezoBridge contract that returns events based on the provided filters. It's up to us to call it with the correct filters. If we want only finalized events, we need to set its `endBlock` field with the finalized block number (64-96 blocks behind the current tip).

ethereum/contracts.go
tomaszslabon: It's the same situation as in https://github.com/mezo-org/mezod/pull/506#discussion_r2261139091.
We shouldn't change anything until bindings for MezoBridgeV2 are available.
Once that happen, we will simply remove this method.

ethereum/sidecar/server.go
tomaszslabon: Done in 12108804b86b4f85a3fbf8e33f80eaf3e4188ac4.
I added the `mezotime` package as you suggested.

ethereum/sidecar/server.go
tomaszslabon: Done in 656493ffbcd5198b430ff2ce59ffdfb7bf4a1d22.
Added more information to the description.

ethereum/contracts.go
tomaszslabon: > Why it will be generated with this name?
Names of these functions are generated in bindings according to names of events in the smart contract, e.g.:
for 
```
event AssetsLocked {...}
```
we have 
```
func (mb *MezoBridge) PastAssetsLockedEvents(
	startBlock uint64,
	endBlock *uint64,
	sequenceNumberFilter []*big.Int,
	recipientFilter []common.Address,
	tokenFilter []common.Address,
) ([]*abi.MezoBridgeAssetsLocked, error) {..}
```

ethereum/contracts.go
lukasz-zimnoch: Ok, thanks for clarifying.

ethereum/contracts.go
lukasz-zimnoch: oh yeah, you are totally right. I had a brain fog and totally missed the fact the name of the event itself is `AssetsUnlockConfirmed` 🤦 Nevermind. Ignore my original comment please.

ethereum/sidecar/server.go
```diff
@@ -434,6 +581,387 @@ func (s *Server) startGRPCServer(
...
+			err,
+		)
+	}
+
+	// The search range can be limited to avoid excessive chain data usage.
+	// When defining the start block we must make sure we cover the entire range
+	// in which the input events could have been confirmed on Ethereum.
+	startBlock := uint64(0)
+	if currentBlock > assetsUnlockConfirmedLookBackBlocks {
+		startBlock = currentBlock - assetsUnlockConfirmedLookBackBlocks
```
lukasz-zimnoch, line 766: Shouldn't we count back from finalized block? What's the point of doing so from current block?

tomaszslabon: Done in 7ec4a2fca2268a922f4759ada90a2ab1144acd2f.
My thinking was: search for `AssetsUnlockConfirmed` events over the last (i.e. since the current tip) ~1 month worth of blocks. But since the difference between the current and finalized blocks is at most 20 minutes, it doesn't matter.

lukasz-zimnoch (approved):
Tested locally to the possible extent (no real `MezoBridge` available yet). Works good!

PR #509, opened 2025-07-03
Implement `AssetsBridge.bridgeOut` method

jeremyletang:
close: https://linear.app/thesis-co/issue/TET-1180/assetsbridge-precompile-the-bridgeout-method

### Introduction

This PR implement the bridgeOut method. This method is the entry point to bridge out from the mezo network ERC20 and BTC back to ethereum.

### Changes

Adds the bridgeOut method in the AssetBridge, this method will in turn:
- verify the inputs from the users
- verify that the bridge is allowed to spend the funds  (either via bank module or evm calls).
- execute the transfer (burnFrom method)
- then calls the x/bridge package to save the AssetsUnlocked event

### Testing

Unit tests have been implemented and all passes, a later PR will implement an extensive suit of system tests for this.

### Author's checklist

### Reviewer's checklist


precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+	// else we do not have enough funds and we need
+	// to return an error
+	if amount.Cmp(balance) > 0 {
+		return fmt.Errorf("not enough funds to bridgeOut")
+	}
+
+	return nil
+}
+
+func (m *BridgeOutMethod) isValidToken(sdkCtx sdk.Context, token common.Address) bool {
```
lukasz-zimnoch, line 417: This function misses the fact that tokens must be validated in the context of the chain (see https://linear.app/thesis-co/document/technical-architecture-9fbdf2ac111c#heading-validation-logic-16c41930). TLDR:
- If Ethereum: BTC and enabled ERC20 are valid tokens
- If Bitcoin: Only BTC is valid

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+}
+
+func (m *BridgeOutMethod) isAmountSpendableByBridge(
+	sdkCtx sdk.Context,
+	spenderAddr common.Address,
+	token common.Address,
+	amount *big.Int,
+) error {
+	// we use the ERC20 allowance call for either the BTC token
+	// or the ERC20 as both implements the ERC20 interface
```
lukasz-zimnoch, line 320:
```suggestion
	// We use the ERC20 allowance call for both the BTC token
	// and ERC20s as both implement the ERC20 interface.
```

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+	return m.isAmountAvailable(sdkCtx, spenderAddr, inputs.Token, inputs.Amount)
+}
+
+func (m *BridgeOutMethod) isAmountSpendableByBridge(
+	sdkCtx sdk.Context,
+	spenderAddr common.Address,
+	token common.Address,
+	amount *big.Int,
+) error {
+	// we use the ERC20 allowance call for either the BTC token
```
lukasz-zimnoch, line 319: Are we sure there are no weird implications we need to consider by calling the precompile this way?

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+		err           error
+		assetUnlocked *bridgetypes.AssetUnlockedEvent
+	)
+
+	switch inputs.Chain {
+	case TargetChainEthereum:
+		assetUnlocked, err = m.executeEthereum(context, inputs)
+	case TargetChainBitcoin:
+		assetUnlocked, err = m.executeBitcoin(context, inputs)
+	}
```
lukasz-zimnoch, line 103: BTC can be bridged out to [both Ethereum (as TBTC) and Bitcoin](https://linear.app/thesis-co/document/technical-architecture-9fbdf2ac111c#heading-validation-logic-16c41930). This condition does not account for this fact.

precompile/assetsbridge/IAssetsBridge.sol
```diff
@@ -43,6 +43,24 @@ interface IAssetsBridge {
...
+    /**
+     * @notice Emitted when an existing asset are unlocked from the bridge.
+     * @param from The address bridging out.
+     * @param recipient The address it's bridged out to on the target chain.
+     * @param token The address of the ERC20 token on the target chain.
+     * @param amount The amount bridged out.
+     * @param chain The chain to which the funds are being bridged out to.
+     * @param sequenceNumber the sequence number for the specific AssetUnlocked.
+     */
+    event AssetUnlocked(
```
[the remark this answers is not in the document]
jeremyletang: ah yea made a mistake here, pretty sure it's AssetsUnlocked down the line.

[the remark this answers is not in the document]
jeremyletang: fixed in: ba5386200865ff0d790ae70872a3e708ab5a2fd9

precompile/assetsbridge/IAssetsBridge.sol
```diff
@@ -43,6 +43,24 @@ interface IAssetsBridge {
...
+        address indexed from,
+        bytes indexed recipient,
```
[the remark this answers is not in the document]
jeremyletang: Sounds good to me. will change.

precompile/assetsbridge/bridge_out.go
jeremyletang: I intend to figure it out by the time I implement the system tests. The only thing I can think of is the performance loss of goign through the EVM execution so far, but I would also think it's negligeable. 

I did that at first in order to keep as much as the flow to be the same, it might not be desirable and I'm happy to change it if necessary.

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+
+func (m *BridgeOutMethod) validateRecipientForChain(chain TargetChain, recipient []byte, rawInput interface{}) error {
+	if len(recipient) == 0 {
+		return fmt.Errorf("recipient can't be empty: %v", rawInput)
+	}
+
+	switch chain {
+	case TargetChainEthereum:
+		// here we just check the length
+		if len(recipient) != 20 {
```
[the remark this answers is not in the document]
jeremyletang: I believe we check earlier if recipient is not 0.

precompile/assetsbridge/IAssetsBridge.sol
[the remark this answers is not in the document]
jeremyletang: Changed in: 5bd199b01a6629fd811a88a7485e0f3807a7131d

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+		}
+	}
+
+	// Check allowed list if it exists
+	// It shouldn't be set seeing that we don't really use cosmos-sdk
+	// but just in case?
+	if len(sendAuth.AllowList) > 0 {
+		found := false
+		senderAddrStr := sdk.AccAddress(common.HexToAddress(
+			evmtypes.AssetsBridgePrecompileAddress,
```
tomaszslabon, line 276: Should sender address be derived from `AssetsBridgePrecompile`?
If so, could you please add some explanation why it is so?

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+//nolint:gosec
+const BridgeOutMethodName = "bridgeOut"
+
+// SendMsgURL defines the authorization type for MsgSend
+var SendMsgURL = sdk.MsgTypeURL(&banktypes.MsgSend{})
+
+// BridgeOutMethod is the implementation of the bridgeOut method.
+type BridgeOutMethod struct {
+	bridgeKeeper BridgeKeeper
+	bankKeeper   BankKeeper
```
tomaszslabon, line 33: `bankKeeper ` is unused. Probably not needed.

tomaszslabon (changes requested):

x/bridge/keeper/assets_unlocked.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	if sequenceTip.IsNil() {
+		sequenceTip = math.ZeroInt()
+	}
+
+	return sequenceTip
+}
+
+// SetAssetsUnlockedSequenceTip sets the current sequence tip for the AssetsLocked
```
tomaszslabon, line 35:
```suggestion
// SetAssetsUnlockedSequenceTip sets the current sequence tip for the AssetsUnlocked
```

precompile/assetsbridge/bridge_out.go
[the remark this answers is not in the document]
lukasz-zimnoch: yeah, let's just add a comment mentioning this fact here

ethereum/sidecar/cli/ethereum_sidecar.go
```diff
@@ -44,6 +45,12 @@ func NewEthereumSidecarCmd() *cobra.Command {
 			defaultServerRequestsPerMinute,
 		))
 
+	cmd.Flags().String(
```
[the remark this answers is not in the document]
jeremyletang: addresed in: 91dccca6

precompile/assetsbridge/IAssetsBridge.sol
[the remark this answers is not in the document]
jeremyletang: fixed in: 31a9016a

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+// - recipient (indexed): the address to which the tokens are transferred
+// - amount (non-indexed): the amount of tokens transferred
+// - chain (non-indexed): the destination chain
+// - sequenceNumber (non-indexed): the sequenceNumber of this AssetsUnlocked
+// - token (non-indexed): the token being bridged out.
+type AssetsUnlockedEvent struct {
+	from, token    common.Address
+	recipient      []byte
+	chain          uint8
+	sequenceNumber *big.Int
```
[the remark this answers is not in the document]
jeremyletang: fixed in: f6dfee4c

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+}
+
+func extractAllowanceFromEVMResult(retData []byte) (*big.Int, error) {
+	if len(retData) != 32 {
+		return nil, fmt.Errorf("invalid return data length")
+	}
+
+	balance := new(big.Int).SetBytes(retData)
+	return balance, nil
+}
```
lukasz-zimnoch, line 615: Seems like both functions can be generalized as `extractUint256` or so.

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+			Indexed: false,
+			Value:   te.chain,
+		},
+	}
+}
+
+// helpers for extracting data out of the EVM response
+
+func extractBalanceOfFromEVMResult(retData []byte) (*big.Int, error) {
+	if len(retData) != 32 {
```
lukasz-zimnoch, line 600: Did we confirm this is actually so simple and the return data is just a 32-byte int that can be directly used to feed a big.Int?

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+	amount *big.Int,
+) error {
+	// We use the ERC20 allowance call for both the BTC token
+	// and ERC20s as both implement the ERC20 interface.
+
+	bridgeAddrBytes := common.HexToAddress(
+		evmtypes.AssetsBridgePrecompileAddress,
+	).Bytes()
+
+	call, err := evmtypes.NewERC20AllowanceCall(
```
lukasz-zimnoch, line 326: Not sure why we need `from` and `to` here. `allowance` takes only two args: `owner` and `spender`.

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+	context *precompile.RunContext,
+	inputs *bridgeOutInputs,
+) error {
+	sdkCtx := context.SdkCtx()
+	// first check that the token is either BTC or a supported ERC20
+	if !m.isValidToken(sdkCtx, inputs.Token) {
+		return fmt.Errorf("unsupported token: %v", inputs.Token)
+	}
+
+	spenderAddr := context.MsgSender()
```
lukasz-zimnoch, line 305: The address calling this method is the owner of the balance. The spender is the `AssetsBridge` precompile.

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+	if err := m.isAmountSpendableByBridge(sdkCtx, spenderAddr, inputs.Token, inputs.Amount); err != nil {
+		return err
+	}
+
+	return m.isAmountAvailable(sdkCtx, spenderAddr, inputs.Token, inputs.Amount)
+}
+
+func (m *BridgeOutMethod) isAmountSpendableByBridge(
+	sdkCtx sdk.Context,
+	spenderAddr common.Address,
```
lukasz-zimnoch, line 315: As mentioned in previous comment:
```suggestion
	owner common.Address,
```

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+	if !m.isValidToken(sdkCtx, inputs.Token) {
+		return fmt.Errorf("unsupported token: %v", inputs.Token)
+	}
+
+	spenderAddr := context.MsgSender()
+	if err := m.isAmountSpendableByBridge(sdkCtx, spenderAddr, inputs.Token, inputs.Amount); err != nil {
+		return err
+	}
+
+	return m.isAmountAvailable(sdkCtx, spenderAddr, inputs.Token, inputs.Amount)
```
lukasz-zimnoch, line 310: Do we actually need to check allowance and balance before burning? For ERC20, the `burnFrom` call does those both checks underneath. Same for BTC as we are relying on the `x/auth` module for allowance and `x/bank` for burning and balance checking. I know the requirement of validating this was mentioned in the technical spec but I think implicit validation done by `burnFrom`/`x/auth`/`x/bank` totally fullfills it. WDYT? This will greatly simplify our code.

x/bridge/keeper/assets_unlocked.go
```diff
@@ -0,0 +1,117 @@
...
+		Sequence:  k.GetAssetsUnlockedSequenceTip(ctx),
+		Recipient: recipient,
+		Amount:    amount,
+		Token:     targetToken.Hex(),
+		Chain:     uint32(chain),
+	}
+
+	k.saveAssetsUnlocked(ctx, assetUnlocked)
+	k.setAssetsUnlockedSequenceTip(
+		ctx, assetUnlocked.Sequence.AddRaw(1),
```
tomaszslabon, line 113: Just a suggestion.
Saving the sequence tip like this will result in the tip always be ahead by `1` compared to the saved `AssetsUnlockedEvent` with the highest `Sequence` so far (the last one saved).

We could increment the tip in such a way as the sequence tip always matches the last `AssetsUnlockedEvent`'s Sequence. This way we would stay consistent with how we [increment sequences in MezoBridge](https://github.com/thesis/mezo-portal/blob/7520174ed2d9e11833c74263f049a91045e4cac8/solidity/contracts/MezoBridge.sol#L181).
@lukasz-zimnoch WDYT?

lukasz-zimnoch: Yeah, I agree with Tomasz's suggestion.

precompile/assetsbridge/bridge_out.go
jeremyletang: Happy to remove the explicit check, I like having those explicitly covered but am also happy to completely remove them if you want.

precompile/assetsbridge/IAssetsBridge.sol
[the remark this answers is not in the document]
jeremyletang: fixed in: 42696a5f

precompile/assetsbridge/IAssetsBridge.sol
```diff
@@ -102,4 +120,17 @@ interface IAssetsBridge {
      * @return The current assets locked sequence tip of the bridge.
      */
     function getCurrentSequenceTip() external view returns (uint256);
+
+    /**
+     * @notice Instructs the bridge to withdraw funds for a give recipient.
+     * @param token The address of the ERC20 token on Mezo.
+     * @param amount The amount of the ERC20 token to withdraw.
+     * @param chain The target chain to withdraw to.
```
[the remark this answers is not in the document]
jeremyletang: fixed in: d023fee6

proto/mezo/bridge/v1/genesis.proto
```diff
@@ -35,4 +35,12 @@ message GenesisState {
     (gogoproto.customtype) = "cosmossdk.io/math.Int",
     (gogoproto.nullable) = false
   ];
+
+  // assets_unlocked_sequence_tip is the current sequence tip for the AssetUnlocked
```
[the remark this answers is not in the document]
jeremyletang: 5516f26e

proto/mezo/bridge/v1/bridge.proto
```diff
@@ -40,6 +40,28 @@ message AssetsLockedEvent {
...
+message AssetUnlockedEvent {
+  // sequence is the unique identifier of the event.
+  string sequence = 1 [
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false
+  ];
+  // recipient is the account address to receive the locked assets on the
+  // target chain.
+  bytes recipient = 2;
+  // amount of assets locked, in 1e18 precision.
```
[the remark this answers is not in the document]
jeremyletang: fixed in: 5516f26e

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+	case TargetChainEthereum:
+		// here we just check the length
+		if len(recipient) != 20 {
+			return fmt.Errorf("invalid recipient address format for Ethereum chain: %v", rawInput)
+		}
+	case TargetChainBitcoin:
+		_, err := btcutil.DecodeAddress(string(recipient), &chaincfg.MainNetParams)
+		if err != nil {
+			return fmt.Errorf("invalid recipient address format for Bitcoin (%v): %v", err, rawInput)
+		}
```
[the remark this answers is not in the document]
jeremyletang: fixed in: 5516f26e

precompile/assetsbridge/bridge_out.go
jeremyletang: fixed: 5516f26e

precompile/assetsbridge/bridge_out.go
jeremyletang: fixed in: cab86f65

precompile/assetsbridge/bridge_out.go
jeremyletang: fixed in: [55ec2b8](https://github.com/mezo-org/mezod/pull/509/commits/55ec2b8dd3d3b61309227bb790196f4d958f46a1)

x/bridge/types/keys.go
```diff
@@ -35,10 +39,21 @@ var (
...
+	AssetsUnlockedSequenceKeyPrefix = []byte{0x80}
 )
 
 // GetERC20TokenMappingKey gets the key for an ERC20 token mapping by the
 // corresponding source token address.
 func GetERC20TokenMappingKey(sourceERC20Token []byte) []byte {
 	return append(ERC20TokenMappingKeyPrefix, sourceERC20Token...)
 }
+
+// GetAssetUnlockKey gets the key for an AssetsUnlocked event.
```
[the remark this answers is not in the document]
jeremyletang: done in: 66d2d505

precompile/assetsbridge/bridge_out.go
lukasz-zimnoch: What are the benefits justifying the explicit checks? I see this may allow us to fail fast but other than that? Do you see anything else?

Regarding the downsides, the major one is more code and increased complexity so more places where we can do mistakes.

Sure, we need to trust that `burnFrom` and `x/bank` will actually check the allowance and balance but this is an inherent part of their logic and not doing so would mean the given ERC20 token or the `x/bank` error have a critical flaw and even if we check explicitly for allowance and balance, that wouldn't help much. Also, if we would be writing this code in pure Solidity, we probably wouldn't do those checks explicitly but just rely on `burnFrom`.

precompile/assetsbridge/bridge_out.go
lukasz-zimnoch: You're right. I missed the fact this is required to implement the `ContractCall` interface. My bad. Nevermind.

precompile/assetsbridge/IAssetsBridge.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Still 3 results sir: 
image

precompile/assetsbridge/bridge_out.go
lukasz-zimnoch: May be soon irrelevant given https://github.com/mezo-org/mezod/pull/509#discussion_r2209898610

x/bridge/keeper/assets_unlocked.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (k Keeper) AssetsUnlocked(
+	ctx sdk.Context,
+	token []byte,
+	amount math.Int,
+	chain uint8,
+	recipient []byte,
+) (*types.AssetsUnlockedEvent, error) {
+	var targetToken common.Address
```
lukasz-zimnoch, line 84: Let's ensure `x/bridge` module doesn't directly depend on `go-ethereum`. Ideally, only `x/evm` and `precompile` packages should. This helps isolating the `go-ethereum` dependency. There are some utils in `x/evm` that `x/bridge` can use to deal with addresses: https://github.com/mezo-org/mezod/blob/cf4a4bc8ffcf9e7c46826c236b26b211921efcb5/x/evm/types/utils.go#L122. Let's leverage that and try to keep the depenendcy graph as `x/bridge` -> `x/evm` -> `go-ethereum`

x/bridge/types/keys.go
```diff
@@ -35,10 +39,21 @@ var (
 	BTCMintedKey = []byte{0x50}
 
 	BTCBurntKey = []byte{0x60}
+
+	// AssetsUnlockedSequenceTipKey is a standalone key for the assets unlocked sequence tip.
+	AssetsUnlockedSequenceTipKey = []byte{0x70}
+
+	// AssetsUnlockedSequenceKeyPrefix is the key prefix for the assets unlocked sequence.
+	AssetsUnlockedSequenceKeyPrefix = []byte{0x80}
```
lukasz-zimnoch, line 47: I would drop the "sequence" part. This is actually just the prefix of the key. Full key = prefix + sequence. 
```suggestion
	// AssetsUnlockedKeyPrefix is the prefix for the assets unlocked key.
	AssetsUnlockedKeyPrefix = []byte{0x80}
```
That would match `GetAssetsUnlockedKey` which then builds the key using `AssetsUnlockedKeyPrefix` and the given `sequence`.

precompile/assetsbridge/bridge_out.go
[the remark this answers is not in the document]
lukasz-zimnoch: I think this is still a bit wrong:
```
		{
			Indexed: false,
			Value:   te.unlockSequenceNumber,
		},
```
This should be indexed. Let's double check carefully.

precompile/assetsbridge/bridge_out.go
jeremyletang: very likely, this should do: https://github.com/mezo-org/mezod/pull/509/commits/36e365dadef62b890227d618604c9df45fba7c72

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,620 @@
...
+func (m *BridgeOutMethod) validateRecipientForChain(chain TargetChain, recipient []byte) error {
+	if len(recipient) == 0 {
+		return fmt.Errorf("recipient can't be empty")
+	}
+
+	switch chain {
+	case TargetChainEthereum:
+		// here we just check the length, the zero address have been tested before already
+		if len(recipient) != 20 {
+			return fmt.Errorf("invalid recipient address format for Ethereum chain: %v", recipient)
```
tomaszslabon, line 508: `recipient` is `[]byte`. We should probably encode it to hex before printing:
```
hex.EncodeToString(recipient)
```
We should do the same when printing for Bitcoin.

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,620 @@
...
+
+	return assetsUnlocked, nil
+}
+
+func (m *BridgeOutMethod) validateAuthorizationLimits(
+	sendAuth *banktypes.SendAuthorization,
+	requestedCoins sdk.Coins,
+) error {
+	if sendAuth.SpendLimit == nil || sendAuth.SpendLimit.Empty() {
+		return fmt.Errorf("no allowance for for %v", requestedCoins[0].Denom)
```
tomaszslabon, line 258:
```suggestion
		return fmt.Errorf("no allowance for %v", requestedCoins[0].Denom)
```

proto/mezo/bridge/v1/bridge.proto
```diff
@@ -40,6 +40,28 @@ message AssetsLockedEvent {
...
 
+// AssetsUnlockedEvent represents the event where inbound assets are released
+// from the bridge.
+message AssetsUnlockedEvent {
+  // sequence is the unique identifier of the event.
+  string unlockSequence = 1 [
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false
+  ];
+  // recipient is the account address to receive the locked assets on the
```
tomaszslabon, line 51:
```suggestion
  // recipient is the account address to receive the unlocked assets on the
```

proto/mezo/bridge/v1/bridge.proto
```diff
@@ -40,6 +40,28 @@ message AssetsLockedEvent {
...
+  // target chain.
+  bytes recipient = 2;
+  // amount of assets locked.
```
tomaszslabon, line 54:
```suggestion
  // amount of assets unlocked.
```

x/bridge/keeper/assets_unlocked.go
```diff
@@ -0,0 +1,117 @@
...
+	} else {
+		if mapping, ok := k.GetERC20TokenMapping(ctx, token); ok {
+			targetToken = common.BytesToAddress(
+				mapping.SourceTokenBytes(),
+			)
+		}
+	}
+
+	if len(targetToken) == 0 {
+		return nil, fmt.Errorf("unknown token %v", token)
```
tomaszslabon, line 100: `token` is `[]bytes`. We should probably encode it to hex before printing:
```
hex.EncodeToString(token)
```
Otherwise, it will be printed as an array of decimal numbers (`[106 20 117 30 ...]`)

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,620 @@
...
+}
+
+func (te *AssetsUnlockedEvent) EventName() string {
+	return AssetsUnlockedEventName
+}
+
+func (te *AssetsUnlockedEvent) Arguments() []*precompile.EventArgument {
+	return []*precompile.EventArgument{
+		{
+			Indexed: false,
```
tomaszslabon, line 585:
```suggestion
			Indexed: true,
```

jeremyletang: 12842c7c

x/evm/types/call.go
```diff
@@ -84,3 +84,180 @@ func (c *ERC20MintCall) To() *common.Address {
...
+
+func (c *ERC20AllowanceCall) To() *common.Address {
+	return &c.to
+}
+
+func (c *ERC20AllowanceCall) Data() []byte {
+	return c.data
+}
+
+// ERC20AllowanceCall represents a burnFrom(address,uin256) call for an ERC20 contract.
```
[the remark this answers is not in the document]
jeremyletang: fa0e60ac

x/bridge/keeper/assets_unlocked.go
jeremyletang: can do, I avoided to do it for "consistency" as for example with recipient it would not always be valid. Not sure it's a good excuse though.

proto/mezo/bridge/v1/bridge.proto
jeremyletang: 0ce0998f

precompile/assetsbridge/bridge_out.go
jeremyletang: 35231d1e, not I haven't done anything for btc yet, as I'm not sure hex makes sense  in that case? Does it?

precompile/assetsbridge/bridge_out.go
jeremyletang: 35231d1e

x/bridge/types/keys.go
jeremyletang: 8e4ba3d0

precompile/assetsbridge/bridge_out.go
jeremyletang: Sounds good, just removing all these.

x/bridge/keeper/assets_unlocked.go
tomaszslabon: We should probably print all incorrect byte data the same way - either print it in decimal format everywhere or hex-encode it everywhere.
Maybe we could print it everywhere as:
```
return nil, fmt.Errorf("unknown token; hex-encoded value: 0x%x", token)
```
This way we will always know we have incorrect data and we print it in hex-encoded format.
@jeremyletang @lukasz-zimnoch 
WDYT?

precompile/assetsbridge/bridge_out.go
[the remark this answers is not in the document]
jeremyletang: b1b1f3aa

precompile/assetsbridge/bridge_out.go
tomaszslabon: Please see https://github.com/mezo-org/mezod/pull/509#discussion_r2222094217.

precompile/assetsbridge/bridge_out.go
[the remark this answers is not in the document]
jeremyletang: This was added as part of another old commit I believe.

x/bridge/keeper/assets_unlocked.go
jeremyletang: done here: 6dc7c4ba

x/bridge/keeper/assets_unlocked.go
jeremyletang: see: 63dc7fd2, not that it's starting at -1 in genesis, so when we increment it, and insert it for the first time the this will match with the actual unlockSequence of the event.

precompile/assetsbridge/bridge_out.go
jeremyletang: done in: 6dc7c4ba5d7999e2ccde4d2358889881cae99aab

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,443 @@
...
+
+// AssetsUnlockedEventName is the name of the AssetsUnlocked event. It matches the name
+// of the event in the contract ABI.
+const AssetsUnlockedEventName = "AssetsUnlocked"
+
+// AssetsUnlockedEvent is the implementation of the AssetsUnlocked event that contains
+// the following arguments:
+// - unlockSequenceNumber (indexed): the sequenceNumber of this AssetsUnlocked
+// - recipient (indexed): the address to which the tokens are transferred
+// - token (non-indexed): the token being bridged out.
```
tomaszslabon, line 382:
```suggestion
// - token (indexed): the token being bridged out.
```

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,443 @@
...
+// - sender (indexed): the address from which the tokens are bridged out,
```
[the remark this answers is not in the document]
jeremyletang: b70a325ebcb15f7e09dd044f920be0cf6532f1eb

precompile/assetsbridge/bridge_out.go
jeremyletang: b70a325ebcb15f7e09dd044f920be0cf6532f1eb

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+	// is it a registered ERC20?
+	if _, ok := m.bridgeKeeper.GetERC20TokenMapping(sdkCtx, token.Bytes()); ok {
+		return true
+	}
+
+	return false
+}
+
+// extractInputs extract the inputs from the precompile.MethodInputs and apply
+// basic validation on the inputs
```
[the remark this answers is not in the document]
jeremyletang: e4593a8a

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,443 @@
...
+
+	assetsUnlocked, err := m.bridgeKeeper.AssetsUnlocked(
+		sdkCtx,
+		inputs.Token.Bytes(),
+		sdkAmount,
+		uint8(inputs.Chain),
+		inputs.Recipient,
+	)
+	if err != nil {
+		// TODO(JEREMY): error happened here, funds might have been
```
tomaszslabon, line 170: What do we do about this TODO?
Are we going to address it in this or the following TODO?

jeremyletang: that's something I'd say is open for discussion still. I would panic to be fair. if you can't write it down to the state, then something v wrong would have happened, and I'd panic to investigate. but I guess it'd be good for you guys to feedback on what you think.

x/bridge/types/genesis.go
```diff
@@ -15,11 +15,12 @@ import (
...
 func DefaultGenesis() *GenesisState {
 	return &GenesisState{
-		Params:                  DefaultParams(),
-		AssetsLockedSequenceTip: sdkmath.NewInt(0),
-		SourceBtcToken:          "",
-		Erc20TokensMappings:     nil,
-		InitialBtcSupply:        sdkmath.NewInt(0),
+		Params:                    DefaultParams(),
+		AssetsLockedSequenceTip:   sdkmath.NewInt(0),
+		AssetsUnlockedSequenceTip: sdkmath.NewInt(-1),
```
tomaszslabon, line 20: I think we should set the tip to `0` initially.
I know this will start numbering `AssetsUnlocked` events from `1` and not `0` (the rip is incremented before being used to assign sequence number for the event), but that's how it works in the `MezoBridge` contract:
- we declare [sequence](https://github.com/thesis/mezo-portal/blob/f2c30f6649c6e63dbec48ab97958a17c95691eea/solidity/contracts/MezoBridge.sol#L18) and its initialized with `0`.
- we [pre-increment the sequence](https://github.com/thesis/mezo-portal/blob/f2c30f6649c6e63dbec48ab97958a17c95691eea/solidity/contracts/MezoBridge.sol#L65) before assigning its value to the event.

This has an advantage: the current value of the tip tells us how many `AssetsUnlock` we have processed. Initially `0`, then `1`, `2`, etc.

tomaszslabon: @jeremyletang 
Could we also have a check for a negative value during genesis, similar to [check of the assets unlocked tip](https://github.com/mezo-org/mezod/blob/b8419881a9a4ae7165dae5912a4ac69f1d3f8555/x/bridge/types/genesis.go#L35)

jeremyletang: sure thing.

x/bridge/types/keys.go
```diff
@@ -35,10 +39,21 @@ var (
...
+	AssetsUnlockedKeyPrefix = []byte{0x80}
 )
 
 // GetERC20TokenMappingKey gets the key for an ERC20 token mapping by the
 // corresponding source token address.
 func GetERC20TokenMappingKey(sourceERC20Token []byte) []byte {
 	return append(ERC20TokenMappingKeyPrefix, sourceERC20Token...)
 }
+
+// GetAssetUnlockedKey gets the key for an AssetsUnlocked event.
```
tomaszslabon, line 56:
```suggestion
// GetAssetsUnlockedKey gets the key for an AssetsUnlocked event.
```

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,446 @@
...
+	recipient []byte,
+	chain uint8,
+	sequenceNumber *big.Int,
+	amount *big.Int,
+) *AssetsUnlockedEvent {
+	return &AssetsUnlockedEvent{
+		sender:               from,
+		recipient:            recipient,
+		token:                token,
+		unlockSequenceNumber: sequenceNumber,
```
tomaszslabon, line 409: Suggestion to consider: passing arguments to `NewAssetsUnlockedEvent` and filling the fields in  `AssetsUnlockedEvent` could be done in the same order as they are declared in  
```
type AssetsUnlockedEvent struct { ... }
```
i.e.
```
	unlockSequenceNumber *big.Int
	recipient            []byte
	token                common.Address
	sender               common.Address
	amount               *big.Int
	chain                uint8
```

precompile/assetsbridge/bridge_out.go
jeremyletang: removed the comment, and keeping as is then. 5451bf62

precompile/assetsbridge/IAssetsBridge.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Bump on this one sir.

proto/mezo/bridge/v1/bridge.proto
```diff
@@ -40,10 +40,33 @@ message AssetsLockedEvent {
...
+message AssetsUnlockedEvent {
+  // unlock_sequence is the unique identifier of the event.
+  string unlock_sequence = 1 [
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false
+  ];
+  // recipient is the account address to receive the unlocked assets on the
+  // target chain.
+  bytes recipient = 2;
+  // amount of assets unlocked.
```
lukasz-zimnoch, line 54: Let's mention the token-specific precision explicitly:
```suggestion
  // amount of assets unlocked, in token-specific precision.
```

x/bridge/keeper/assets_unlocked.go
lukasz-zimnoch: But we actually want it to be 0 at genesis and 1 for the first entry. Can we change that?

precompile/assetsbridge/bridge_out.go
lukasz-zimnoch: I see this was addressed. Closing.

precompile/assetsbridge/bridge_out.go
lukasz-zimnoch: +1 on this

precompile/assetsbridge/bridge_out.go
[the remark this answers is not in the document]
lukasz-zimnoch: This was addressed. Closing.

precompile/assetsbridge/bridge_out.go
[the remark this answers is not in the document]
lukasz-zimnoch: Bump on this sir.

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,615 @@
...
+		return fmt.Errorf("failed to execute ERC20 allowance call: %w", err)
+	}
+
+	// this would be non empty if the call reverted
+	if len(resp.VmError) > 0 {
+		// TODO(Jeremy): Should we handle the revert properly
+		// to return an error message? For now just returning
+		// a generic error message
+		return fmt.Errorf("ERC20 allowance call reverted")
+	}
```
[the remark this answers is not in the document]
lukasz-zimnoch: Bump on this sir.

precompile/assetsbridge/bridge_out.go
jeremyletang: done in: 6f6f0faf

precompile/assetsbridge/bridge_out.go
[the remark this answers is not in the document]
jeremyletang: oh dear, that might be a wrong hash as well. let me find the correct one.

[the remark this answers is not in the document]
jeremyletang: This for the change: 92a50f3e8b8b3ad58f571d9b86e29e544f0abf2f
This for the relevant tests: 891c1c46d7766ff97d6c59386941bc98af7cf41e

The commit I linked initially was actually the one where I fixed the tests for the validation of the btc addressees

precompile/assetsbridge/bridge_out.go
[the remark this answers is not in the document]
jeremyletang: fix have been removed before as part of removing all the extra validation, whcih should have been linked in another comment, let me find it back.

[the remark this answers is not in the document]
jeremyletang: 891c1c46d7766ff97d6c59386941bc98af7cf41e should be it.

precompile/assetsbridge/IAssetsBridge.sol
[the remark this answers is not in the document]
jeremyletang: yea I've just checked again and I cannot find any of them remaining, I might have addressed them as I saw them. Unless you find some more, but I just grep-ed the whole codebase for `assetUnlock AssetUnlock`  and couldn't find any.

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,443 @@
...
+	if expiration != nil && expiration.Before(context.SdkCtx().BlockTime()) {
+		return nil, fmt.Errorf("authorization expired at %v", expiration)
+	}
+
+	sdkAmount, err := precompile.TypesConverter.BigInt.ToSDK(inputs.Amount)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert amount: [%w]", err)
+	}
+	coins := sdk.Coins{{Denom: evmtypes.DefaultEVMDenom, Amount: sdkAmount}}
+	msg := banktypes.NewMsgSend(bridgeAddrBytes, senderAddr.Bytes(), coins)
```
tomaszslabon, line 203: I find the `executeBitcoin` function quite hard to follow.
For example, it seems like in this line:
```
msg := banktypes.NewMsgSend(bridgeAddrBytes, senderAddr.Bytes(), coins)
```
we are sending funds from the Bridge to the sender (the caller of the `bridgeOut` method).
So the sender gains funds. Shouldn't they lose funds if they are requesting bridging-out?

Short descriptions for `executeEthereum` and `executeBitcoin` would be helpful. For example they could briefly explain what the functions do with the focus on how the funds are moved/burned.

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,443 @@
...
+
+	_, err = m.authzKeeper.DispatchActions(context.SdkCtx(), bridgeAddr, []sdk.Msg{msg})
```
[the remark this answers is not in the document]
jeremyletang: you're correct, I was actually implementing just that as I was re-reading it.

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,516 @@
...
+	resp, err := sendAuth.Accept(context.SdkCtx(), msg)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't update authorization: %w", err)
+	}
+
+	if resp.Updated != nil {
+		err = m.authzKeeper.SaveGrant(context.SdkCtx(), bridgeAddr, senderAddr, resp.Updated, expiration)
+	} else {
+		// Authorization fully consumed, delete it
+		m.authzKeeper.DeleteGrant(context.SdkCtx(), bridgeAddr, senderAddr, SendMsgURL)
```
lukasz-zimnoch, line 232: Let's remember that the same authorization instance may hold approvals for the MEZO token. Let's double check the grant is NOT deleted in case BTC allowance is zero but MEZO is not. We should delete the grant only if all tokens/coins are zero.

precompile/hardhat/contracts/AssetsBridgeCaller.sol
```diff
@@ -37,4 +41,5 @@ contract AssetsBridgeCaller is IAssetsBridge {
     function getSourceBTCToken() external view returns (address) {
         return IAssetsBridge(precompile).getSourceBTCToken();
     }
+
```
lukasz-zimnoch, line 44: Nit: Probably redundant new line.

proto/mezo/bridge/v1/bridge.proto
```diff
@@ -31,7 +31,7 @@ message AssetsLockedEvent {
   // recipient is the account address to receive the locked assets on Mezo,
   // in Bech32 format.
   string recipient = 2;
-  // amount of assets locked, in 1e18 precision.
+  // amount of assets locked.
```
lukasz-zimnoch, line 34: Good we are fixing that but let's add "in token-specific precision" instead.

x/bridge/types/keys.go
```diff
@@ -35,10 +39,21 @@ var (
...
 )
 
 // GetERC20TokenMappingKey gets the key for an ERC20 token mapping by the
 // corresponding source token address.
 func GetERC20TokenMappingKey(sourceERC20Token []byte) []byte {
 	return append(ERC20TokenMappingKeyPrefix, sourceERC20Token...)
 }
+
+// GetAssetsUnlockedKey gets the key for an AssetsUnlocked event.
+func GetAssetsUnlockedKey(sequence math.Int) []byte {
```
lukasz-zimnoch, line 57: Nit regarding consistency:
```suggestion
func GetAssetsUnlockedKey(unlockSequence math.Int) []byte {
```

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,516 @@
...
+		).Bytes()
+		spenderAddr = context.MsgSender()
+	)
+
+	sdkAmount, err := precompile.TypesConverter.BigInt.ToSDK(inputs.Amount)
+	if err != nil {
+		return nil, fmt.Errorf("unable to convert amount to sdk type: %v", err)
+	}
+
+	call, err := evmtypes.NewERC20BurnFromCall(
```
lukasz-zimnoch, line 147: Thoughts on moving the burn call to `x/bridge` and expose it as `BurnERC20` from the keeper? We do that for Bitcoin. Also, minting happens entirely within `x/bridge`. We would have a consistent approach were the precompile is just an interface and the `x/bridge` module is responsible for the logic.

x/bridge/keeper/assets_unlocked.go
```diff
@@ -0,0 +1,146 @@
...
+		Amount:         amount,
+		Token:          targetToken,
+		Chain:          uint32(chain),
+	}
+	k.saveAssetsUnlocked(ctx, assetsUnlocked)
+
+	return assetsUnlocked, nil
+}
+
+func (k Keeper) BurnBTC(
```
lukasz-zimnoch, line 123: Thing to consider: we could also move the authorization management from the precompile here and rename this method to `BurnBTCFrom`. I don't have a strong preference. Just throwing an idea.

x/bridge/keeper/assets_unlocked.go
```diff
@@ -0,0 +1,146 @@
...
+) (*types.AssetsUnlockedEvent, error) {
+	var targetToken string
+	// is it the btc token?
+	btcToken := evmtypes.HexAddressToBytes(
+		evmtypes.BTCTokenPrecompileAddress,
+	)
+	if bytes.Equal(btcToken, token) {
+		targetToken = evmtypes.BytesToHexAddress(k.GetSourceBTCToken(ctx))
+	} else {
+		if mapping, ok := k.GetERC20TokenMapping(ctx, token); ok {
```
lukasz-zimnoch, line 91: `GetERC20TokenMapping` expects the SOURCE token as input, not the MEZO token. We need to search from the other side.

proto/mezo/bridge/v1/bridge.proto
```diff
@@ -40,10 +40,33 @@ message AssetsLockedEvent {
   string token = 4;
 }
 
-// ERC20TokenMapping defines a mapping between an ERC20 token on the source chain
-// and on the Mezo chain.
+// AssetsUnlockedEvent represents the event where inbound assets are released
+// from the bridge.
+message AssetsUnlockedEvent {
```
lukasz-zimnoch, line 45: Let's please reorder fields in the same way as in `IAssetsBridge`. This will be a help for readers. Remember about updating usage places (`SaveAssetsUnlocked`).

proto/mezo/bridge/v1/bridge.proto
```diff
@@ -40,10 +40,33 @@ message AssetsLockedEvent {
   string token = 4;
 }
 
-// ERC20TokenMapping defines a mapping between an ERC20 token on the source chain
-// and on the Mezo chain.
+// AssetsUnlockedEvent represents the event where inbound assets are released
+// from the bridge.
```
lukasz-zimnoch, line 44: I would also consider storing the `sender` just in case we need to implement bridge out timeouts one day. We need to know where to return funds.

x/evm/types/call.go
```diff
@@ -84,3 +84,180 @@ func (c *ERC20MintCall) To() *common.Address {
...
+func (c *ERC20BalanceOfCall) To() *common.Address {
+	return &c.to
+}
+
+func (c *ERC20BalanceOfCall) Data() []byte {
+	return c.data
+}
+
+// ERC20AllowanceCall represents a allowance(address,address) call for an ERC20 contract.
+type ERC20AllowanceCall struct {
```
lukasz-zimnoch, line 148: Unused code. Let's remove.

x/evm/types/call.go
```diff
@@ -84,3 +84,180 @@ func (c *ERC20MintCall) To() *common.Address {
 func (c *ERC20MintCall) Data() []byte {
 	return c.data
 }
+
+// ERC20BalanceOfCall represents a balanceOf(address) call for an ERC20 contract.
+type ERC20BalanceOfCall struct {
```
lukasz-zimnoch, line 89: Unused call. Let's get rid of it.

x/bridge/keeper/assets_unlocked.go
```diff
@@ -0,0 +1,146 @@
+package keeper
```
lukasz-zimnoch, line 1: Let's make sure all those methods are covered by unit tests.

precompile/assetsbridge/assets_bridge.go
```diff
@@ -169,4 +212,36 @@ type BridgeKeeper interface {
...
+}
+
+type EvmKeeper interface {
+	// ExecuteContractCall executes an EVM contract call.
+	ExecuteContractCall(
+		ctx sdk.Context,
+		call evmtypes.ContractCall,
+	) (*evmtypes.MsgEthereumTxResponse, error)
+	// IsContract returns if the account contains contract code.
+	IsContract(ctx sdk.Context, address []byte) bool
```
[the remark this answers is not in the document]
jeremyletang: yea i had planned to use it before, but not in the end: bb01e034

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,516 @@
...
+	}
+
+	switch chain {
+	case TargetChainEthereum:
+		// here we just check the length, the zero address have been tested before already
+		if len(recipient) != 20 {
+			return fmt.Errorf("invalid recipient address format for Ethereum chain: %v", hex.EncodeToString(recipient))
+		}
+	case TargetChainBitcoin:
+		if keepbtc.GetScriptType(recipient) == keepbtc.NonStandardScript {
```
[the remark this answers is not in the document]
jeremyletang: Ah fair point, I thought the implementation was inside the keep repository when I looked into it.

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,516 @@
...
+	}
+
+	if err := m.validateRecipientForChain(inputs.Chain, inputs.Recipient); err != nil {
+		return err
+	}
+
+	// first check that the token is either BTC or a supported ERC20
+	if !m.isValidToken(sdkCtx, inputs.Token, inputs.Chain) {
+		return fmt.Errorf("unsupported token: %v", inputs.Token)
+	}
```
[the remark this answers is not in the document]
jeremyletang: ac8cbcfa

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,516 @@
...
+				assetsUnlocked.Amount.BigInt(),
+			),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to emit AssetsUnlocked event: [%w]", err)
+		}
+
+	}
+
+	return precompile.MethodOutputs{err == nil}, err
```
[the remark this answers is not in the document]
jeremyletang: 71e9c437

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,516 @@
...
+	chain                uint8
+}
+
+func NewAssetsUnlockedEvent(
+	sequenceNumber *big.Int,
+	recipient []byte,
+	token common.Address,
+	sender common.Address,
+	chain uint8,
+	amount *big.Int,
```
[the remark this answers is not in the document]
jeremyletang: 4e59617b

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,516 @@
...
+		spenderAddr = context.MsgSender()
+	)
+
+	sdkAmount, err := precompile.TypesConverter.BigInt.ToSDK(inputs.Amount)
+	if err != nil {
+		return nil, fmt.Errorf("unable to convert amount to sdk type: %v", err)
+	}
+
+	call, err := evmtypes.NewERC20BurnFromCall(
+		bridgeAddrBytes,
```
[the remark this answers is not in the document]
jeremyletang: 1c6b60ea

x/evm/types/call.go
```diff
@@ -84,3 +84,180 @@ func (c *ERC20MintCall) To() *common.Address {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create uint256 type: %w", err)
+	}
+
+	methodAbi := abi.Method{
+		Name: "burnFrom",
+		ID:   []byte{0x79, 0xcc, 0x67, 0x90}, // 0x79cc6790 is the function selector for burnFrom(address,uint256)
+		Type: abi.Function,
+		Inputs: []abi.Argument{
+			{Name: "address", Type: addressType},
```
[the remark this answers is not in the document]
jeremyletang: ba594759

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,516 @@
...
+	authorization, expiration := m.authzKeeper.GetAuthorization(
+		context.SdkCtx(), bridgeAddr.Bytes(), senderAddr.Bytes(), SendMsgURL,
+	)
+	if authorization == nil {
+		return nil, fmt.Errorf("%s authorization type does not exist or is expired for address %s", SendMsgURL, senderAddr)
+	}
+
+	if expiration != nil && expiration.Before(context.SdkCtx().BlockTime()) {
+		return nil, fmt.Errorf("authorization expired at %v", expiration)
+	}
```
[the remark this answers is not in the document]
jeremyletang: a98cf24f

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,516 @@
...
+	if err := m.validate(context, inputs); err != nil {
+		return precompile.MethodOutputs{false}, err
+	}
+
+	return m.execute(context, inputs)
+}
+
+// execute will execute the burn of the token then send
+// AssetsUnlocked events to the bridgeKeeper.
+func (m *BridgeOutMethod) execute(
```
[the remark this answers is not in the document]
jeremyletang: b7dcb115

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,516 @@
...
+	// burnt funds
+	journal := context.Journal()
+	journal.SubBalance(common.BytesToAddress(senderAddr.Bytes()), balanceDelta, tracing.BalanceChangeTransfer)
+
+	return assetsUnlocked, nil
+}
+
+func (m *BridgeOutMethod) validateAuthorizationLimits(
+	sendAuth *banktypes.SendAuthorization,
+	requestedCoins sdk.Coins,
```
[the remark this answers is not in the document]
jeremyletang: 1888d3cf

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,516 @@
...
+		context.SdkCtx(),
+		senderAddr.Bytes(),
+		sdkAmount,
+	); err != nil {
+		return nil, fmt.Errorf("couldn't burn BTC: %w", err)
+	}
+
+	// now update the authorization to spend for the AssetsBridge
+	msg := banktypes.NewMsgSend(senderAddr.Bytes(), bridgeAddr.Bytes(), coins)
+	resp, err := sendAuth.Accept(context.SdkCtx(), msg)
```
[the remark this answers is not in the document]
jeremyletang: 9e9a8421

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,516 @@
...
+
+		if requestedCoin.Amount.GT(allowedAmount) {
+			return fmt.Errorf(
+				"requested amount %s exceeds allowed amount %s for %s",
+				requestedCoin.Amount,
+				allowedAmount,
+				requestedCoin.Denom,
+			)
+		}
+	}
```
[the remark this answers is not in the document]
jeremyletang: 9e9a8421

x/bridge/types/keys.go
jeremyletang: 4f27495f

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,443 @@
...
+	case TargetChainEthereum:
+		if _, ok := m.bridgeKeeper.GetERC20TokenMapping(sdkCtx, token.Bytes()); ok {
+			return nil
+		}
+
+		return fmt.Errorf("unsupported token: %v for ethereum target chain", token)
+	case TargetChainBitcoin:
+		return fmt.Errorf("unsupported token: %v for bitcoin target chain", token)
+	default:
+		panic("unreachable")
```
tomaszslabon, line 280: Nitpick, but maybe a more precise message:
```suggestion
		panic("unreachable: unknown chain type")
```

proto/mezo/bridge/v1/bridge.proto
jeremyletang: 4bf631d5

proto/mezo/bridge/v1/bridge.proto
```diff
@@ -40,10 +40,35 @@ message AssetsLockedEvent {
...
+  // token is the hex-encoded EVM address of the bridged token.
+  string token = 3;
+  // sender is the address bridging out.
+  bytes sender = 4;
+  // amount of assets unlocked, in token-specific precision.
+  string amount = 5 [
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false
+  ];
+  // chain is the target chain for this unlock event
```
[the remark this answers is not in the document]
jeremyletang: e8458689

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,443 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to convert amount: [%w]", err)
+	}
+	coin := sdk.Coin{Denom: evmtypes.DefaultEVMDenom, Amount: sdkAmount}
+
+	// now update the authorization to spend for the AssetsBridge
+	msg := banktypes.NewMsgSend(senderAddr.Bytes(), bridgeAddr.Bytes(), sdk.NewCoins(coin))
+	resp, err := sendAuth.Accept(context.SdkCtx(), msg)
+	if err != nil {
+		return fmt.Errorf("couldn't update authorization: %w", err)
```
lukasz-zimnoch, line 186: Looks like we have a small mistake here:
```suggestion
		return fmt.Errorf("couldn't check authorization: %w", err)
```

precompile/assetsbridge/bridge_out.go
```diff
@@ -0,0 +1,443 @@
...
+	if resp.Delete {
+		// Authorization fully consumed, delete it
+		err = m.authzKeeper.DeleteGrant(context.SdkCtx(), bridgeAddr, senderAddr, SendMsgURL)
+	} else if resp.Updated != nil {
+		err = m.authzKeeper.SaveGrant(context.SdkCtx(), bridgeAddr, senderAddr, resp.Updated, expiration)
+	}
+
+	if err != nil {
+		return fmt.Errorf("bridge is not authorized to burn BTC: %w", err)
+	}
```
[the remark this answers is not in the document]
jeremyletang: d24a4d67

lukasz-zimnoch (approved):

PR #527, opened 2025-08-12
Expose keyring in ethereum sidecar

jeremyletang:
close: https://linear.app/thesis-co/issue/TET-1185/ethereum-sidecar-access-to-operator-private-key

### Introduction

Give access to the keyring to the ethereum-sidecar

### Changes

This PRs adds the flags to the ethereum-sidecar related to the keyring.
It also moves the extractPrivateKey private to client/keys/utils and make it public to be accessible elsewhere
It recover the keys inside of the ethereum-sidecar at startup (this later can be changed when the next step are implemented to only load them at the time of usage just so the key do not stay loaded in memory at all time).

### Testing

Initialise a localnode `make localnode-bin-start`.
Then run the ethereum-sidecar:
```
mezod ethereum-sidecar --key-name=dev0 --keyring-backend=test --keyring-type=eth_secp256k1 --home=.localnode
```
You should see a log stating that the private keys have been loaded
```
9:44AM INF Successfully extracted private key from keyring key-name=dev0 module=ethereum-sidecar

```
---

### Author's checklist

- [ ] Provided the appropriate description of the pull request

### Reviewer's checklist


ethereum/sidecar/cli/flags.go
```diff
@@ -12,6 +12,10 @@ const (
 	FlagServerNetwork             = "ethereum-sidecar.server.network"
 	FlagServerBatchSize           = "ethereum-sidecar.server.batch-size"
 	FlagServerRequestsPerMinute   = "ethereum-sidecar.server.requests-per-minute"
+	FlagKeyringBackend            = "keyring-backend"
+	FlagKeyringDir                = "keyring-dir"
+	FlagKeyType                   = "keyring-type"
```
lukasz-zimnoch, line 17: We support only ECDSA keys in the sidecar so there is probably no point of exposing this flag. Let's remove it and make sure we fail if non-ECDSA key is loaded upon sidecar startup.

go.sum
lukasz-zimnoch, line 1: Why we modify go.sum?

PR #546, opened 2025-08-27
Add bridge worker endpoint

jeremyletang:
Closes: https://linear.app/thesis-co/issue/TET-1193/bridge-worker-attestation-endpoint

### Introduction

Adds and HTTP server for the bridge-worker, the server for now have a single endpoint which:
- receive new attestation and signature bundles
- validate both (valid attestation on mezo, valid sequence number, authorized signer, etc)
- and store both in supabase if the validation was successful.

### Changes

This introduce an HTTP server in the bridge-worker package.
For conveniences a Client has been added, interface used by the sidecar batch attestation have been updated.

### Testing

This is covered with unit test, to run them:
```
go test -v ./bridge-worker
```

### Author's checklist

### Reviewer's checklist


bridge-worker/server.go
```diff
@@ -0,0 +1,195 @@
...
+func writeSuccess(w http.ResponseWriter, status int) {
+	w.WriteHeader(status)
+	_ = json.NewEncoder(w).Encode(types.SubmitSignatureResponse{
+		Success: true,
+	})
+}
+
+// abiEncodeAttestationWithChainID is used to encode the attestation with the chain ID
+// which is used to produce a signature for the batch attestation process.
+func abiEncodeAttestationWithChainID(attestation *types.AssetsUnlocked, chainID *big.Int) ([]byte, error) {
```
lukasz-zimnoch, line 149: Let's not duplicate this method from the sidecar. Instead, let's extract it into `ethereum/bindings/portal` and make it rely on `*portal.MezoBridgeAssetsUnlocked`. This package feels right as this method is about encoding of a binding type.

bridge-worker/server.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (s *Server) recoverAddress(entry *types.AssetsUnlocked, signature string) (common.Address, error) {
+	abiEncoded, err := abiEncodeAttestationWithChainID(entry, s.chainID)
+	if err != nil {
+		return common.Address{}, err
+	}
+
+	// Apply the same TextHash transformation
+	hash := accounts.TextHash(abiEncoded)
```
lukasz-zimnoch, line 99: This is wrong, we need to keccak256 the `abiEncoded` var before passing to `TextHash`. I think best way to ensure both ends of the signature are consistent is extracting `attestationDigestHash` to a common place as well. Let's also move it to `ethereum/bindings/portal`.

bridge-worker/types/types.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	if err := validateSignature(s.Signature); err != nil {
+		return err
+	}
+
+	return s.Entry.Validate()
+}
+
+type SubmitSignatureResponse struct {
```
lukasz-zimnoch, line 97:
```suggestion
type SubmitAttestationResponse struct {
```

bridge-worker/types/types.go
```diff
@@ -0,0 +1,119 @@
...
+	if !strings.HasPrefix(signature, "0x") {
+		return ErrInvalidSignatureFormat
+	}
+
+	signature = signature[2:]
+
+	bytes, err := hex.DecodeString(signature)
+	if err != nil {
+		return ErrInvalidSignatureFormat
+	}
```
lukasz-zimnoch, line 112: This can be probably replaced with `hexutil.Decode`

bridge-worker/server_test.go
```diff
@@ -0,0 +1,351 @@
...
+			} else {
+				assert.Equal(t, w.Code, testCase.expectedStatus)
+				assert.False(t, response.Success)
+				assert.Contains(t, response.Error, testCase.expectedError)
+			}
+		})
+	}
+}
+
+func TestServer_recoverAddress(t *testing.T) {
```
lukasz-zimnoch, line 314: I believe we should provide proper test vectors for this function and treat the tested function as a black box. This test must ensure the signature will be accepted by the contract. Right now, we are just taking the logic of the tested function, and reverse it to produce the test vector. This will always pass. See that we are missing a keccak256 that is checked on the contract side and this test still pass. Let's use existing Solidity tests to get a test vector for this test: https://github.com/thesis/mezo-portal/blob/47fcf2dd33c681c9ef41df2b8ff470b5ef4fab47/solidity/test/MezoBridge.test.ts#L209

bridge-worker/server.go
```diff
@@ -0,0 +1,195 @@
...
+		Success: false,
+	})
+}
+
+func writeSuccess(w http.ResponseWriter, status int) {
+	w.WriteHeader(status)
+	_ = json.NewEncoder(w).Encode(types.SubmitSignatureResponse{
+		Success: true,
+	})
+}
```
lukasz-zimnoch, line 145: Those are specific to our endpoint, let's either make them private functions inside the handler or extend their names to contain the endpoint name. We will likely have other endpoints in the future here, hence my point.

ethereum/sidecar/batch_attestation.go
```diff
@@ -12,22 +12,23 @@ import (
...
+	retrySubmitSignature    = 5 * time.Second
 
 	ErrBridgeWorkerNotSet = errors.New("bridge worker not set")
 )
 
 type BridgeWorker interface {
 	// expect no returned payload,
 	// just an error eventually
-	SendSignature(attestation *portal.MezoBridgeAssetsUnlocked, signature string) error
+	SubmitSignature(attestation *bwtypes.AssetsUnlocked, signature string) error
```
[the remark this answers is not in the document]
jeremyletang: 240193e6

bridge-worker/server.go
```diff
@@ -0,0 +1,195 @@
...
+	port uint16,
+	chainID *big.Int,
+) *Server {
+	s := &Server{
+		logger:  logger,
+		chainID: chainID,
+	}
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("POST /submit-signature", s.submitSignature)
```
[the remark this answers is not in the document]
jeremyletang: 240193e6

bridge-worker/server.go
```diff
@@ -0,0 +1,195 @@
...
+	publicKeyBytes, err := crypto.SigToPub(hash, signatureBytes)
+	if err != nil {
+		return common.Address{}, err
+	}
+
+	// Convert public key to address
+	return crypto.PubkeyToAddress(*publicKeyBytes), nil
+}
+
+func readSubmitSignatureRequest(r *http.Request) (*types.SubmitSignatureRequest, error) {
```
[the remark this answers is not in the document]
jeremyletang: 240193e6

bridge-worker/types/types.go
jeremyletang: 240193e6

bridge-worker/types/types.go
```diff
@@ -0,0 +1,119 @@
...
+	ErrInvalidRecipient           = errors.New("invalid recipient")
+	ErrInvalidToken               = errors.New("invalid token")
+	ErrMissingAmount              = errors.New("missing amount")
+	ErrInvalidAmount              = errors.New("invalid amount")
+	ErrInvalidChain               = errors.New("invalid chain")
+
+	zero = big.NewInt(0)
+)
+
+type AssetsUnlocked struct {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yes those are valid points but I think we need to look from a higher perspective: we have `x/bridge` which is our core component defining the domain model and two other components - the sidecar and the bridge worker - that communicate between themselves. From this perspective, both sidecar and bridge component should directly depend only on the `x/bridge`, not on their own types. That said, `x/bridge/types.AssetsUnlocked` feels like a natural choice. Even if we don't need all fields or we need to do some conversions, this seems to be an acceptable price as we avoid introducing additional types and keep the higher-level dependency graph clean.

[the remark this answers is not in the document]
jeremyletang: 58afc4f938dc3aa7bef5007c9d79a4a20658a77e

ethereum/sidecar/batch_attestation.go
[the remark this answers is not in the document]
jeremyletang: e4a3e9c248a7ac73450170c3946fee48a11330a9

bridge-worker/server.go
jeremyletang: 97d098942c681bf7e6256cc93df2aa9c84b6ad0d

bridge-worker/server.go
jeremyletang: 69c5cc3d22996cd9ecce0663bbf8899e924f4491

bridge-worker/bridge_worker.go
```diff
@@ -185,8 +202,19 @@ func RunBridgeWorker(
 		}(job)
 	}
 
+	go func() {
+		defer cancelCtx()
+		env.server.Start()
+		env.logger.Warn("Http server stopped")
```
lukasz-zimnoch, line 208: Nit: Let's use lowercase for log messages:
```suggestion
		env.logger.Warn("http server stopped")
```

bridge-worker/server.go
```diff
@@ -0,0 +1,253 @@
...
+	isConfirmed, err := s.mezoBridge.ConfirmedUnlocks(req.Entry.UnlockSequence.BigInt())
+	if err != nil {
+		s.logger.Error("couldn't confirm unlock", "error", err)
+		s.writeSubmitAttestationError(w, err, http.StatusInternalServerError)
+		return
+	}
+
+	// if already confirmed, nothing to do
+	if isConfirmed {
+		s.writeSubmitAttestationError(w, errors.New("already confirmed"), http.StatusBadRequest)
```
lukasz-zimnoch, line 155: This is not an error. Can happen from time to time if there are time drifts and a honest validator submits its signature with a delay. I would return a 200 in this case.

bridge-worker/server.go
```diff
@@ -0,0 +1,253 @@
...
+		return
+	}
+
+	// finally, is it a valid attestation
```
lukasz-zimnoch, line 159:
```suggestion
	// finally, check attestation validity

File: bridge-worker/server.go [hunk 14]
```diff
@@ -0,0 +1,253 @@
...
+	publicKeyBytes, err := crypto.SigToPub(hash, signatureBytes)
+	if err != nil {
+		return common.Address{}, err
+	}
+
+	// Convert public key to address
+	return crypto.PubkeyToAddress(*publicKeyBytes), nil
+}
+
+func toPortalAssetsUnlock(entry *bridgetypes.AssetsUnlockedEvent) *portal.MezoBridgeAssetsUnlocked {
```
lukasz-zimnoch, line 229: Maybe:
```suggestion
func toAssetsUnlockedBindingType(entry *bridgetypes.AssetsUnlockedEvent) *portal.MezoBridgeAssetsUnlocked {
```

bridge-worker/server.go
```diff
@@ -0,0 +1,253 @@
...
+	// finally, is it a valid attestation
+	attestation := toPortalAssetsUnlock(req.Entry)
+	ok, err := s.mezoBridge.ValidateAssetsUnlocked(*attestation)
+	if err != nil {
+		s.logger.Error("couldn't validate assets unlock event", "error", err)
+		s.writeSubmitAttestationError(w, err, http.StatusInternalServerError)
+		return
+	}
+
+	// if already confirmed, nothing to do
```
lukasz-zimnoch, line 168: This case means the entry is invalid. Let's fix this comment.

bridge-worker/server.go
```diff
@@ -0,0 +1,253 @@
...
+	if !ok {
+		s.writeSubmitAttestationError(w, errors.New("not a valide asset unlocked event"), http.StatusBadRequest)
```
lukasz-zimnoch, line 170:
```suggestion
		s.writeSubmitAttestationError(w, errors.New("not a valid AssetsUnlocked event"), http.StatusBadRequest)
```

bridge-worker/store.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"cosmossdk.io/log"
+	"cosmossdk.io/math"
+	bridgetypes "github.com/mezo-org/mezod/x/bridge/types"
+	"github.com/supabase-community/supabase-go"
+)
+
+const (
+	AssetsUnlockedEventsTable  = "assets_unlocked_events"
+	AttestationSignaturesTable = "attestations_signatures"
```
lukasz-zimnoch, line 14: `attestations_signatures` doesn't match the table name from the SQL migration

bridge-worker/server.go
```diff
@@ -0,0 +1,253 @@
...
+		return
+	}
+
```
lukasz-zimnoch, line 173: We also need to make sure the given validator cannot submit more than one signature for the given unlock sequence. We should probably rely on address not on the validator ID during this check as IDs may change and be re-assigned upon validator set changes.

bridge-worker/store.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// SaveAttestation will save an attestation in the database
+// because there's no good atomic way of checking the error that
+// occurred, we take an optimistic approach, meaning:
+// we first try to insert
+// if we have an error we then try to get the entry
+// if we are successful at loading the entry, then we return no error
+// if we are not then we return the initial error
+func (s *SupabaseStore) SaveAttestation(entry *bridgetypes.AssetsUnlockedEvent) error {
```
lukasz-zimnoch, line 45: What if the first validator submits malicious entry data (e.g. changes the recipient)? The rest of validators will be stuck as the worker will get wrong data as the state of the world. I think we need to model the behavior as follows:
1. Accept valid `entry` (unlock sequence + data set containing amount, recipient, etc.) + `signature` pairs
2. Assume each unlock sequence may have multiple different data sets
3. Count the number of backing signatures behind each data set
4. The data set with quorum of signatures becomes canonical for the given unlock sequence and is submitted by the worker

This is basically how the `MezoBridge` contract works for individual attestations and I think we need to do the same in the Bridge Worker.

cc @tomaszslabon as this is also relevant for the batch attestation job you're working on.

bridge-worker/server.go
```diff
@@ -0,0 +1,253 @@
...
+		s.logger.Error("couldn't save assets unlock event", "error", err)
+		s.writeSubmitAttestationError(w, err, http.StatusInternalServerError)
+		return
+	}
+
+	if err := s.store.SaveSignature(req.Entry.UnlockSequence, req.Signature); err != nil {
+		s.logger.Error("couldn't save signature", "error", err)
+		s.writeSubmitAttestationError(w, err, http.StatusInternalServerError)
+		return
+	}
```
lukasz-zimnoch, line 185: I think we should aim to execute the entry and signature save as a single SQL command. Additionally, we should double-check our data was properly stored inside Supabase. This is a cloud database and we already observed weird corner cases in the past so better safe than sorry.

bridge-worker/store.go
```diff
@@ -0,0 +1,122 @@
...
+	return &SupabaseStore{
+		logger: logger,
+		client: client,
+	}, nil
+}
+
+// SaveAttestation will save an attestation in the database
+// because there's no good atomic way of checking the error that
+// occurred, we take an optimistic approach, meaning:
+// we first try to insert
```
lukasz-zimnoch, line 41: If we know saving duplicated data will error out, I think it's better to try load the data before save and skip in case the data already exist.

bridge-worker/config.go
```diff
@@ -37,6 +37,11 @@ type ConfigProperties struct {
 
 	JobBTCWithdrawalQueueCheckFrequency time.Duration
 
+	ServerPort uint16
+
+	SupabaseURL string
+	SupbaseKey  string
```
tomaszslabon, line 43:
```suggestion
	SupabaseKey  string
```

cmd/bridge-worker/flags.go
```diff
@@ -108,6 +114,18 @@ func newFlagSet() *flag.FlagSet {
...
 	)
 
+	fs.String(
+		flagSupabaseURL,
+		"",
+		"The Supabase URL",
+	)
+
+	fs.Uint16(
+		flagSupabaseURL,
```
tomaszslabon, line 124:
```suggestion
		flagHTTPServerPort,
```

supabase/migrations/20250922191246_initial_tables.sql
```diff
@@ -0,0 +1,16 @@
...
+  recipient character varying not null,
+  token character varying not null,
+  sender character varying not null,
+  amount character varying not null,
+  chain int not null,
+  block_time numeric not null,
+  constraint assets_unlocked_events_pkey primary key (unlock_sequence)
+);
+
+create table signatures (
```
tomaszslabon, line 12: Probably:
```suggestion
create table attestations_signatures (
```
as in the Go code we have:
```
	AttestationSignaturesTable = "attestations_signatures"
```

bridge-worker/client/client.go
```diff
@@ -0,0 +1,57 @@
...
+
+type Client struct {
+	baseURL    string
+	httpClient *http.Client
+}
+
+func NewClient(baseURL string) *Client {
+	return &Client{
+		baseURL:    baseURL,
+		httpClient: &http.Client{},
```
tomaszslabon, line 26: Maybe add a timeout to the HTTP Client?
```
http.Client{Timeout: time.Duration(1) * time.Second}
```

PR #556, opened 2025-09-09
Introduce Bitcoin-specific min amount for bridge out

lukasz-zimnoch:
References: 

### Testing

---

### Author's checklist

- [ ] Provided the appropriate description of the pull request

### Reviewer's checklist


precompile/assetsbridge/min_amount.go
```diff
@@ -207,3 +207,134 @@ func (m *GetMinBridgeOutAmountMethod) Run(
...
+
+	err = m.bridgeKeeper.SetMinBridgeOutAmountForBitcoinChain(
+		context.SdkCtx(),
+		sdkMinAmount,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return precompile.MethodOutputs{true}, nil
```
jeremyletang, line 288: not really a big deal, but why return true at all? there's never a case where false can be returned, the method will always revert on error if not mistaken.

jeremyletang (dismissed):
Tested it locally, works as advertised.

precompile/assetsbridge/min_amount.go
lukasz-zimnoch: It matters where precompile functions are called from regular Solidity contracts. If EVM encounters a method without return value, it calls `extcodesize` to see whether the call is against a contract. However, not all of our precompiles may have the bytecode set and if so, the call won't be executed as `extcodesize` returns 0. Because of that, we developed a standard practice to return `bool` from precompile write methods to not bother with this problem at all.

precompile/assetsbridge/IAssetsBridge.sol
```diff
@@ -213,6 +213,35 @@ interface IAssetsBridge {
     * @notice Returns the minimum bridge-out amount (if set) for a Mezo token.
     * @param mezoToken The address of the token on the Mezo chain.
     * @return The current minimum amount for the token (0 if unset).
+    * @dev For bridge outs to the Bitcoin chain, the system will always check
+    *      the Bitcoin-specific minimum returned by getMinBridgeOutAmountForBitcoinChain
+    *      first, and fallback to the value returned by this function if the
```
tomaszslabon, line 218: I think this description needs updating.
If the target chain is Bitcoin and Bitcoin-specific is non-zero, we check both minimum bridge-out amounts.

precompile/assetsbridge/min_amount.go
```diff
@@ -207,3 +207,134 @@ func (m *GetMinBridgeOutAmountMethod) Run(
...
+
+func (m *SetMinBridgeOutAmountForBitcoinChainMethod) RequiredGas(_ []byte) (uint64, bool) {
+	return 0, false
+}
+
+func (m *SetMinBridgeOutAmountForBitcoinChainMethod) Payable() bool {
+	return false
+}
+
+func (m *SetMinBridgeOutAmountForBitcoinChainMethod) Run(
```
tomaszslabon, line 250: When setting the general minimum bridge-out amount we emit an event.
We could consider emitting an event here for consistency.

precompile/assetsbridge/min_amount.go
jeremyletang: Fair enough

x/bridge/keeper/assets_unlocked.go
```diff
@@ -226,3 +226,33 @@ func (k Keeper) SetMinBridgeOutAmount(
...
+	minAmount math.Int,
+) error {
+	store := ctx.KVStore(k.storeKey)
+	bz, err := minAmount.Marshal()
+	if err != nil {
+		panic(err)
+	}
+
+	store.Set(types.MinBridgeOutAmountForBitcoinChainKey, bz)
+	return nil
```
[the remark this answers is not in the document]
lukasz-zimnoch: Good call, see 4591b1dd9cc29c98b85ed3a604cedbdb3147fa1e

precompile/assetsbridge/min_amount.go
lukasz-zimnoch: See 236dcc18d37c984b05103ee7c69e62fdafee6b5d

tomaszslabon (approved):

PR #651, opened 2026-04-13
Use mezo-org/go-ethereum v1.16.9

tomaszslabon:
Closes: https://linear.app/thesis-co/issue/MEZO-4012/go-124-version-bump
Closes: https://linear.app/thesis-co/issue/MEZO-4009/custom-precompile-adaptation

Note: this PR should be merged into the feature branch `geth-upgrade`.

### Introduction

This PR upgrades the `mezo-org/go-ethereum` library to `1.16.9`.
NOTE: this update aims at preserving the current `mezod` features. It does not implement the new features that come with forks from the new `mezo-org/go-ethereum`. TODOs are left in the code.

### Changes

- Bumped the version of Go to `1.24`
- Used the new `mezo-org/go-ethereum` version in such a way as to preserve already existing features.
- Added a CI job checking if there are TODOs for geth upgrade.

### Testing

Run unit and system tests.
Perform basic manual testing on localnet.

---

### Author's checklist

### Reviewer's checklist


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/mezo-org/mezod
 
-go 1.22.11
+go 1.24.0
```
lukasz-zimnoch, line 3: The `go 1.24.0` bump here is incomplete — these still point at 1.22:

- `Dockerfile:6` — `golang:1.22.11-bullseye`
- `.github/workflows/test.yml:20` — `go-version: '1.22'`
- `.github/workflows/build.yml:18` — `go-version: '1.22'`
- `.github/workflows/lint.yml:20` — `go-version: '1.22'`
- `.github/workflows/binary.yml:23` — `go-version: 1.22`
- `CLAUDE.md` — _"Go toolchain: go 1.22.11 (pinned in go.mod and Dockerfile)"_

CI only passes because `setup-go` defaults to `GOTOOLCHAIN=auto` and pulls 1.24 on the fly. Anyone with `GOTOOLCHAIN=local` will break, and the Docker image is relying on the same fallback. Let's bring all of these in line.

x/evm/statedb/statedb.go
```diff
@@ -394,6 +397,16 @@ func (s *StateDB) Witness() *stateless.Witness {
 	return s.witness
 }
 
+func (s *StateDB) AccessEvents() *gethstate.AccessEvents {
+	// TODO: implement when adding support for the new access-event flow in 1.16.9.
```
lukasz-zimnoch:
```suggestion
	// TODO (geth-upgrade): implement when adding support for the new access-event flow in 1.16.9.
```

Let's tag all the 1.16.9 TODOs with a common `TODO (geth-upgrade):` prefix, and add a temporary CI job that runs only on `geth-upgrade` and PRs targeting it, greps for the marker, and fails if any match. That way we can't forget any of these before merging back to `main`.

Note on the `AccessEvents()` current state: all usages of the returned value in upstream `core/vm/*` are gated by `evm.chainRules.IsEIP4762`, so `nil` is fine for our current fork set. But it's a nil-dereference the moment Osaka/Verkle activates — we need to be careful and re-visit this when the time comes. The tagged TODO should be a good reminder.

x/evm/statedb/statedb.go
```diff
@@ -394,6 +397,16 @@ func (s *StateDB) Witness() *stateless.Witness {
...
+	return nil
+}
+
+//nolint:misspell
+func (s *StateDB) Finalise(bool) {
+	// TODO: implement when adding support for the new state finalization flow in 1.16.9.
```
lukasz-zimnoch:
```suggestion
	// TODO (geth-upgrade): implement when adding support for the new state finalization flow in 1.16.9.
```

Same `TODO (geth-upgrade):` marker — see the `AccessEvents` comment above.

My gut regarding current implementation: `Finalise(bool)` is only called from upstream `core/state_processor.go`. We don't go through `StateProcessor` (our path is `applyMessageWithConfig`), so the no-op is safe. Worth double checking once we have more knowledge at the end of the upgrade work though.

x/evm/statedb/statedb.go
```diff
@@ -439,6 +452,16 @@ func (s *StateDB) GetCommittedState(addr common.Address, hash common.Hash) commo
 	return common.Hash{}
 }
 
+// GetStateAndCommittedState returns the current value and the committed value.
+func (s *StateDB) GetStateAndCommittedState(addr common.Address, hash common.Hash) (common.Hash, common.Hash) {
+	// TODO: check when reviewing state access changes in 1.16.9.
```
lukasz-zimnoch:
```suggestion
	// TODO (geth-upgrade): check when reviewing state access changes in 1.16.9.
```

Same `TODO (geth-upgrade):` marker — see the `AccessEvents` comment above.

x/evm/keeper/fees.go
```diff
@@ -94,7 +94,8 @@ func VerifyFee(
 		accessList = txData.GetAccessList()
 	}
 
-	intrinsicGas, err := core.IntrinsicGas(txData.GetData(), accessList, isContractCreation, homestead, istanbul, isShanghai)
+	// TODO: pass set code authorizations when supporting the new 1.16.9 transaction flow.
```
lukasz-zimnoch:
```suggestion
	// TODO (geth-upgrade): pass set code authorizations when supporting the new 1.16.9 transaction flow.
```

Same `TODO (geth-upgrade):` marker — see the `AccessEvents` comment in `x/evm/statedb/statedb.go`.

x/evm/keeper/gas.go
```diff
@@ -38,7 +38,8 @@ func (k *Keeper) GetEthIntrinsicGas(ctx sdk.Context, msg core.Message, cfg *para
 	istanbul := cfg.IsIstanbul(height)
 	isShanghai := cfg.IsShanghai(height, uint64(time)) //nolint:gosec
 
-	return core.IntrinsicGas(msg.Data, msg.AccessList, isContractCreation, homestead, istanbul, isShanghai)
+	// TODO: pass set code authorizations when supporting the new 1.16.9 transaction flow.
```
lukasz-zimnoch:
```suggestion
	// TODO (geth-upgrade): pass set code authorizations when supporting the new 1.16.9 transaction flow.
```

Same `TODO (geth-upgrade):` marker — see the `AccessEvents` comment in `x/evm/statedb/statedb.go`.

precompile/contract.go
```diff
@@ -96,6 +96,11 @@ func (c *Contract) Address() common.Address {
 	return c.address
 }
 
+// Name returns a stable identifier of the precompile.
+func (c *Contract) Name() string {
+	return c.address.Hex()
+}
```
lukasz-zimnoch, line 104: Out of curiosity, we don't need to change anything: how is this field used in geth? Moreover, why we picked the address and not an arbitrary name?

x/feemarket/keeper/eip1559.go
```diff
@@ -85,11 +85,10 @@ func (k Keeper) CalculateBaseFee(ctx sdk.Context) *big.Int {
...
 		x := new(big.Int).Mul(parentBaseFee, gasUsedDelta)
 		y := x.Div(x, parentGasTargetBig)
-		baseFeeDelta := math.BigMax(
-			x.Div(y, baseFeeChangeDenominator),
-			common.Big1,
-		)
-
+		baseFeeDelta := x.Div(y, baseFeeChangeDenominator)
+		if baseFeeDelta.Cmp(common.Big1) < 0 {
+			return x.Add(parentBaseFee, common.Big1)
```
lukasz-zimnoch: Wouldn't it be cleaner to update `baseFeeDelta` here by setting it to `common.Big1` and then return once  after that?

go.mod
tomaszslabon: Done in e22137e4d7ff41c5274d9b072e3fdcd6f8e0d42f.

x/evm/statedb/statedb.go
tomaszslabon: Updated the TODO and added a CI workflow in e10e6b40ed652999a9db4fa644d8fe4c8f2a6e4b.

x/evm/statedb/statedb.go
tomaszslabon: Updated the TODO in e10e6b40ed652999a9db4fa644d8fe4c8f2a6e4b.

x/evm/statedb/statedb.go
tomaszslabon: Updated the TODO in e10e6b40ed652999a9db4fa644d8fe4c8f2a6e4b.

x/evm/keeper/fees.go
tomaszslabon: Updated the TODO in e10e6b40ed652999a9db4fa644d8fe4c8f2a6e4b.

x/evm/keeper/gas.go
tomaszslabon: Updated the TODO in e10e6b40ed652999a9db4fa644d8fe4c8f2a6e4b.

x/feemarket/keeper/eip1559.go
tomaszslabon: Done in ad9c57b837d84de49984b50b9b06abe72be68358.

precompile/contract.go
tomaszslabon: It seems the only usage for `Name()` is [internal/ethapi/api.go](https://github.com/mezo-org/go-ethereum/blob/v1.16.9-mezo0/internal/ethapi/api.go#L1206). It's not called in mezod.
That's why I used precompile's addresses as their names - to change as little as possible.

However, after thinking about it I decided to implement Name() as string for each precompile in f251bd384ad9b9bae34a447eaaf2c7203a52e2bf.
If we are forced to implement the `Contract.Name()` method, let's at least use a reasonable name rather than just an address. We might use this method in the future.

The names are:
`btc-token`
`mezo-token`
`validator-pool`
`assets-bridge`
`price-oracle`
`maintenance`
`upgrade`

x/feemarket/keeper/eip1559.go
lukasz-zimnoch: I meant:
```
baseFeeDelta := x.Div(y, baseFeeChangeDenominator)
if baseFeeDelta.Cmp(common.Big1) < 0 {
	baseFeeDelta = common.Big1
}
return new(big.Int).Add(parentBaseFee, baseFeeDelta)
```

tomaszslabon: Done in de04ff51aa1782b29b3a5e70f23b0e0c61cedf70.

lukasz-zimnoch (approved):

PR #658, opened 2026-04-21
Scaffold eth_simulateV1 RPC surface

lukasz-zimnoch:
References: [MEZO-4227](https://linear.app/thesis-co/issue/MEZO-4227/implement-eth-simulatev1-if-possible)

### Introduction

This PR is Phase 1 of 12 for `eth_simulateV1` support in mezod (see `.claude/MEZO-4227-eth-simulate-v1/plan.md`). It lands the JSON-RPC method name on the `eth_` namespace with a documented "not yet implemented" error so tooling can discover the method, while the real simulation pipeline ships in subsequent phases on this branch. Zero behavior risk — every code path either short-circuits with a structured error or belongs to new types / tests.

#### New request / response types (`rpc/types/simulate.go`)

Spec-conformant types for the `eth_simulateV1` request / response payloads per [`ethereum/execution-apis`](https://github.com/ethereum/execution-apis/blob/main/src/eth/execute.yaml): `SimOpts`, `SimBlock`, `BlockOverrides`, `SimCallResult`, `SimBlockResult`, `SimCallError`. `BlockOverrides` covers every spec field plus the two extras geth's `override.BlockOverrides` also carries (`Difficulty`, `BeaconRoot`) so later phases can explicitly reject what mezo doesn't support (withdrawals, beacon root). `SimCallResult.MarshalJSON` coerces a nil `Logs` slice to `[]` rather than `null`. `SimBlockResult` uses a custom `MarshalJSON` / `UnmarshalJSON` pair so the block fields sit at the top level of the JSON envelope alongside `calls`.

#### Spec-reserved JSON-RPC error codes (`rpc/types/errors.go`)

A full set of constants for every code the execution-apis spec assigns to `eth_simulateV1`: `-38010`..`-38015`, `-38020`..`-38026`, `-32005`, `-32015`, `-32016`, `-32601`..`-32603`, and `3` for reverted calls (EIP-140 / spec `CallResultFailure`). Each constant carries a doc comment quoting the spec's textual description so future readers can see exactly which condition the code pins to. The `-38022` / `-38023` identifiers (`SimErrCodeMovePrecompileSelfReference`, `SimErrCodeMovePrecompileDuplicateDest`) match the spec text rather than a mezo-internal interpretation. Also exports a generic `RPCError{Code, Message, Data}` type with `Error()` / `ErrorCode()` / `ErrorData()` methods so geth's rpc encoder populates the wire-level `{code, message, data}` fields correctly — not simulate-specific, so future `eth_call` / `eth_estimateGas` / send-tx paths that want structured codes can reuse it rather than falling through to a bare `-32000` error.

#### Backend stub (`rpc/backend/simulate.go` + interface addition in `rpc/backend/backend.go`)

`Backend.SimulateV1(opts, blockNrOrHash)` returns `rpctypes.NewSimNotImplementedError()` — a `-32603` internal error carrying the message `"eth_simulateV1 is not yet implemented"`. `-32603` is the spec-listed internal-error code for this method; it is preferred over `-32601` "method not found" because the method IS registered on the `eth_` namespace and returning -32601 would mislead clients that cache unknown-method signals. `-32601` stays reserved for the operator kill switch landing in Phase 8, which intentionally hides the endpoint wholesale. `SimulateV1` is added to the `EVMBackend` interface alongside `DoCall`.

#### RPC namespace handler (`rpc/namespaces/ethereum/eth/simulate.go` + interface addition in `rpc/namespaces/ethereum/eth/api.go`)

`PublicAPI.SimulateV1` forwards to the backend. go-ethereum's rpc server auto-registers exported methods, so this appears as `eth_simulateV1` in `rpc_modules` output. Added to the `EthereumAPI` interface in the `EVM/Smart Contract Execution` section next to `Call`.

#### Plan file

Full phased plan for the rollout at `.claude/MEZO-4227-eth-simulate-v1/plan.md`, with Phase 1 marked done and the Phase 2 / 5 / 8 / 10 designs pinned to the spec-conformant error codes listed above.

### Testing

- Go unit tests (`rpc/types/simulate_test.go`, `rpc/types/errors_test.go`, `rpc/backend/simulate_test.go`) cover: JSON round-trips for every type, empty-`Logs` coercion, populated logs, unknown-fields tolerance on unmarshal, the full set of error-code constants including `-38012`, `-32016`, and `3`, `RPCError` interface implementation, and the backend stub returning `-32603` with a "not yet implemented" message both under empty opts and under a non-empty `SimOpts` plus explicit `BlockNumberOrHash`.
- Hardhat system test (`tests/system/test/SimulateV1_Stub.test.ts`) calls `provider.send("eth_simulateV1", [{blockStateCalls: []}, "latest"])` and asserts the response surfaces a `-32603` error whose message references `eth_simulatev1` and contains `not yet implemented`.
- `go build ./...` is green; `golangci-lint run` against the touched packages produces no findings on new code.

Running the full unit suite:

```
go test ./rpc/types/ ./rpc/backend/
ok  	github.com/mezo-org/mezod/rpc/types	0.074s
ok  	github.com/mezo-org/mezod/rpc/backend	3.731s
```

---

### Author's checklist

### Reviewer's checklist


rpc/types/errors.go
```diff
@@ -0,0 +1,59 @@
...
+	SimErrCodeClientLimitExceeded         = -38026
+
+	SimErrCodeFeeCapTooLow   = -32005
+	SimErrCodeVMError        = -32015
+	SimErrCodeTimeout        = -32016
+	SimErrCodeMethodNotFound = -32601
+	SimErrCodeInvalidParams  = -32602
+	SimErrCodeInternalError  = -32603
+
+	SimErrCodeReverted = 3
```
tomaszslabon, line 38: `SimErrCodeReverted` is set to `3` while the [research.md](https://github.com/mezo-org/mezod/blob/e09434a6b0abfc0be2b683620c3e674cb8a8e4ab/.claude/MEZO-4227-eth-simulate-v1/research.md?plain=1#L251) describes it as `-32000` for geth.
It also seems to be what geth repo actually uses for this type of [error](https://github.com/mezo-org/go-ethereum/blob/859c41bdcb7141276fc9c7a7c3486190025f9ec1/internal/ethapi/errors.go#L115).

lukasz-zimnoch: This one's deliberate — picked in 84f0c93b. Spec and geth disagree here:

- `execute.yaml`'s `CallResultFailure` schema pins `code: const: 3` and says "error codes are enforced by the spec"
- geth ignores that and reuses its legacy `-32000` from `eth_call` (your link)
- reth follows the spec (`SIMULATE_REVERT_CODE = 3` in `crates/rpc/rpc-eth-types/src/simulate.rs:37`)

The `research.md` line you quoted is me transcribing geth's behavior, not prescribing it — sorry, that reads misleading.

I'd like to keep `3`: it's what the schema enforces, it matches reth, and our schema-driven conformance fixtures won't need a mezo-specific carve-out. Annotated the constant in 6d19dfdb so the next reader doesn't trip on it.

tomaszslabon (approved):

PR #660, opened 2026-04-22
Add MovePrecompileTo support to eth_call state overrides

lukasz-zimnoch:
References: [MEZO-4227](https://linear.app/thesis-co/issue/MEZO-4227/implement-eth-simulatev1-if-possible)
Depends on: https://github.com/mezo-org/mezod/pull/658

### Introduction

This PR is Phase 2 of 12 for `eth_simulateV1` support in mezod (see `.claude/MEZO-4227-eth-simulate-v1/plan.md`). It lands `MovePrecompileTo` support on the state-override machinery that `eth_call` already consumes. The same override type will be reused by `eth_simulateV1` in later phases, so shipping the feature on `eth_call` now gives a validated override path before the simulate driver is wired up. Zero behavior change for callers that do not set the new field; consensus-critical paths (`ApplyTransaction`, `ApplyMessage`, `ApplyMessageWithConfig`, `NewEVM`) are semantically unchanged.

#### `OverrideAccount` gains `movePrecompileToAddress`

`rpc/types/types.go` and the internal `x/evm/keeper.overrideAccount` both acquire a `MovePrecompileTo *common.Address` field using the JSON tag `movePrecompileToAddress` mandated by the [execution-apis spec](https://github.com/ethereum/execution-apis/blob/main/src/eth/execute.yaml). When set, the precompile at that source address is relocated to the destination address for the duration of the call; the source loses its precompile binding, so its regular account fields (`Nonce`, `Code`, `Balance`, `State`, `StateDiff`) may then be overridden in the same request.

Only standard Ethereum precompiles (`0x01`–`0x0A`) are movable. Mezo custom precompiles at `0x7b7c…` (`BTCToken`, `MEZOToken`, `ValidatorPool`, `AssetsBridge`, `Maintenance`, `Upgrade`, `PriceOracle`, and the testbed) are rejected by a denylist derived from `types.DefaultPrecompilesVersions` so the policy cannot drift when a new custom precompile is added.

#### `evmBuilder` seam threads the move through `applyMessageWithConfig`

`applyMessageWithConfig` gains a single `buildEVM evmBuilder` parameter; its body is otherwise byte-identical to main — the only substituted line is `evm := k.NewEVM(...)` → `evm := buildEVM(...)`. The public `ApplyMessageWithConfig` wrapper passes `k.NewEVM` directly, so `ApplyTransaction`, `ApplyMessage`, and every other consensus caller sees identical behavior. `NewEVM` itself is unchanged.

`SimulateMessage` validates the override set up front via `applyStateOverrides` (which returns the collected source → destination map), then hands `applyMessageWithConfig` a closure that wraps `k.NewEVM` and — for requests with at least one move — rebuilds the active precompile registry, applies the relocations, and re-attaches via `evm.WithPrecompiles(...)`. The inline registry rebuild is the only duplication introduced by the feature, marked `TODO (geth-upgrade)` to collapse to `evm.Precompiles()` + `evm.SetPrecompiles()` once the chain is on go-ethereum v1.16.9.

#### `applyStateOverrides` returns the validated move set and domain sentinels

`applyStateOverrides` now returns `(map[common.Address]common.Address, error)` — the collected source → destination relocations — alongside mutating the StateDB for the regular account overrides. Move validation uses `vm.DefaultPrecompiles(rules)` (for "is the source a stdlib precompile?") and `mezoCustomPrecompileAddrs` (for the deny-list), so no precompile registry map is threaded through.

On the first invariant violation the function returns one of the sentinel errors defined in `x/evm/types/override_errors.go` wrapped with the offending address via `fmt.Errorf("%w: %s", sentinel, addr)`. The seven sentinels each name a single invariant checked by the validator:

| Sentinel | Invariant |
|---|---|
| `ErrOverrideMovePrecompileSelfReference` | `movePrecompileToAddress == addr` (spec fatal) |
| `ErrOverrideMovePrecompileDuplicateDest` | Two overrides target the same destination (spec fatal) |
| `ErrOverrideNotAPrecompile` | Source address is not a stdlib precompile |
| `ErrOverrideMoveMezoCustomPrecompile` | Source address is a mezo custom precompile (denylist) |
| `ErrOverrideDestAlreadyOverridden` | `diff.has(dst)` — destination also carries a regular override |
| `ErrOverrideAccountTaintedByPrecompile` | `dirtyAddrs[addr]` — re-overriding an address already taken by a move |
| `ErrOverrideStateAndStateDiff` | Both `state` and `stateDiff` supplied for one account |

The last two guards mirror go-ethereum's `override.go` checks and eliminate Go-map-iteration-order nondeterminism for pathological input like `{A: {MoveTo:B}, B: {MoveTo:C}}`.

Keeper code stays free of the JSON-RPC error vocabulary — `x/evm/keeper` no longer imports `rpc/types` for error construction, and unit tests assert against the sentinels with `require.ErrorIs`.

#### Wire-code surface deferred to the `eth_simulateV1` gRPC path

The spec-reserved JSON-RPC codes (`-38022`, `-38023`, `-32602`) only matter on the `eth_simulateV1` path; `eth_call` has no dedicated codes for these conditions. Translating the sentinels to those codes therefore waits for the `SimulateV1` gRPC handler. The plan's new Error-surface sections in `.claude/MEZO-4227-eth-simulate-v1/plan.md` (Phase 2 and Phase 4) document the chosen design: `SimulateV1Response` will carry a structured `override_error_kind` enum populated in the keeper from these sentinels, and the backend will translate the enum to `*RPCError` at the RPC boundary. `rpc/backend/simulate.go` gets a `TODO` anchor so the wiring intent is visible where the code will land.

The proto-field channel is the shape the existing `res.VmError` / `res.Failed()` VM-error path already uses (`rpc/backend/call_tx.go:482`) and the only shape that preserves structured error information across the `status.Error(codes.Internal, err.Error())` wrapping the shared gRPC query handler applies to every response.

### Testing

- Keeper unit tests at `x/evm/keeper/state_override_test.go` cover: `CollectsMove` (move returned in the map), `SourceNotPrecompile`, `SelfReference`, `DuplicateDestination`, `MezoCustomBlocked` (table-driven over every entry of `DefaultPrecompilesVersions`), `DestAlreadyOverridden`, `ChainedMoveRejected`, `WithSourceCodeOverwrite` (move plus source code/nonce override coexist). Assertions go through `require.ErrorIs` against the `x/evm/types` sentinels so the domain-layer contract does not drift silently. Every pre-existing `TestApplyStateOverrides_*` test continues to pass unchanged.
- Hardhat system test at `tests/system/test/SimulateV1_MovePrecompile_ethCall.test.ts` exercises the full JSON-RPC stack end-to-end: moves `sha256` to `0x…1234` and asserts the returned bytes equal the precomputed `sha256("hello")`; moves the `BTCToken` precompile and asserts the rejection message; moves `sha256` onto itself and asserts the self-reference rejection.
- Manual localnode verification confirmed every rejection path surfaces the expected message text and the success path returns the correct hash.
- `go build ./...` clean; `go test -race ./x/evm/keeper/... ./rpc/types/...` green; `golangci-lint run ./x/evm/keeper/... ./rpc/types/...` produces no findings on new code.

---

### Author's checklist

### Reviewer's checklist


tomaszslabon:
I tested locally that moving standard precompiles during `eth_call` simulation works, and that moving Mezo custom precompiles is rejected as expected.

tomaszslabon (approved):

PR #662, opened 2026-04-23
Single-call/block eth_simulateV1

lukasz-zimnoch:

### Introduction

Ships the first working slice of `eth_simulateV1`: a single simulated block containing a single call, state overrides honored (including `movePrecompileToAddress` for stdlib precompiles), a spec-shaped block envelope in the response, and every spec-reserved JSON-RPC error code reaching the client verbatim. Multi-call, multi-block, `traceTransfers`, `validation=true`, `returnFullTransactions`, DoS guards, and full spec conformance are follow-up work.

The feature is query-side only — no proto changes to `service Msg`, no genesis or stored-state mutations, no ante handler edits, no `ApplyTransaction` edits. `NewEVM` is refactored to delegate to a new `NewEVMWithOverrides` with a nil override pointer, producing byte-identical block context / coinbase / base-fee / random / difficulty against the pre-refactor path. Rolling-upgrade-safe.

### Changes

- **Keeper helpers.** `NewEVMWithOverrides` takes an optional `*EVMOverrides` (block context, precompile registry, base-fee toggle) and reproduces `NewEVM` byte-for-byte when the pointer is nil. `StateDB.FinaliseBetweenCalls` drops per-call ephemeral state (logs, refund, transient storage, precompile-call counter) while preserving state objects, access list, and journal — used by the simulate driver to run calls against a shared StateDB. The synthesized base header inherits `Coinbase` from `cfg.CoinBase` so the `COINBASE` opcode reports the validator operator inside simulate, matching the non-simulate path. When the rpc backend supplies a canonical base block hash on the request, the first simulated block's `parentHash` echoes it instead of hashing the synthetic base header, so the value matches what `eth_getBlockByNumber` surfaces for the same block. Per-call nonce defaults are read from the simulated StateDB so caller-supplied `stateOverrides[from].nonce` values are honored. Default inter-block timestamp gap is 3 seconds, matching mezo's CometBFT cadence.
- **Unified error surface.** `x/evm/types.SimError{Code, Message, Data}` with one `NewSim*` constructor per spec-reserved failure mode. Implements geth's `ErrorCode()`/`ErrorData()` so the JSON-RPC server emits `{code, message, data}` verbatim. Rides the gRPC wire on a dedicated `SimError error = 2` field on `SimulateV1Response`. `applyStateOverrides` returns `*SimError` over the plain `error` channel; callers that do not care about the code propagate it unchanged, callers that do branch with `errors.As`.
- **Simulate surface.** New `SimulateV1` RPC on `ethermint.evm.v1.Query`. Bare types (`SimOpts`, `SimBlock`, `SimBlockOverrides`, `SimCallResult`, `SimBlockResult`) and their helpers (`UnmarshalSimOpts`, `BuildSimCallResult`) live under `x/evm/types`. The driver plus `sanitizeSimChain` (strictly-increasing block numbers/timestamps, gap-fill, span-check before allocation), `makeSimHeader`, `assembleSimBlock`, and `computeSimTxHash` live under `x/evm/keeper`. `StateOverride` and `OverrideAccount` live under `x/evm/types` as the single definition used by both `eth_simulateV1` and `eth_call`/`eth_estimateGas`; the duplicate `rpc/types` copies are removed.
- **RPC backend.** `Backend.SimulateV1` marshals `SimOpts` to JSON, resolves the target block via `BlockNumberFromTendermint` so hash-form `BlockNumberOrHash` inputs pick the correct header, forwards the canonical CometBFT block hash on a new `BaseBlockHash` request field — `tmtypes.Header.Hash()` of the already-fetched block, the same value `FormatBlock` surfaces as `hash` for `eth_getBlockByNumber` — so the keeper can use it for the first simulated block's `parentHash`, forwards over the gRPC query client, and returns `*SimError` verbatim on the error channel so geth picks up the spec code; timeout wiring mirrors `DoCall`.
- **Hardhat system tests.** `tests/system/test/SimulateV1_SingleCall.test.ts` deploys `SimpleToken`, mints, and runs two `eth_simulateV1` calls: a happy-path ERC-20 `transfer` with a sender-balance state override (asserts `status 0x1`, decoded `true` return data, a `Transfer` log with correct topics and amount) and a revert-path transfer that exceeds balance (asserts `status 0x0` plus a per-call error with code `3`, message `"execution reverted"`, and data beginning with the Solidity 0.8 `Panic(0x11)` selector `0x4e487b71`). `tests/system/test/SimulateV1_RejectedOverrides.test.ts` sends one request per unsupported `BlockOverrides` field (`beaconRoot`, `withdrawals`, `blobBaseFee`) and asserts the server returns JSON-RPC `-32602` at request level with a message naming the rejected field. The Phase 1 stub test (`SimulateV1_Stub.test.ts`) is removed — its `-32603 "not yet implemented"` assertion is unconditionally false against the real implementation.

### Testing

- `make test-unit` — passes. New coverage: `TestNewEVMWithOverrides`, `TestFinaliseBetweenCalls`, `TestSimulateV1_*` against `KeeperTestSuite` (empty opts, single-call happy path, `MovePrecompileTo` for stdlib sha256, state-override sentinel bubbles up, nil-request guard, unsupported-EIP override rejection, base-block-hash override of the first simulated block's `parentHash`), white-box tests for `sanitizeSimChain` and `makeSimHeader`, black-box tests for `UnmarshalSimOpts` / `BuildSimCallResult` / `SimBlockResult.MarshalJSON`, `SimError` catalog coverage alongside the constructors, and an rpc-backend test asserting the canonical base hash is forwarded on `SimulateV1Request.BaseBlockHash`.
- `make proto-all` — clean.
- `make lint` — clean on the changed files.
- Hardhat system tests — `./tests/system/system-tests.sh SimulateV1_SingleCall SimulateV1_RejectedOverrides` runs all ten cases green against a fresh `make localnode-bin-start`.

---

### Author's checklist

- [x] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`) — docs update for the method and divergences lands with spec conformance.

### Reviewer's checklist


x/evm/keeper/simulate_v1.go
```diff
@@ -0,0 +1,409 @@
...
+	parent *ethtypes.Header,
+	overrides *types.SimBlockOverrides,
+	rules params.Rules,
+	chainCfg *params.ChainConfig,
+	validation bool,
+) *ethtypes.Header {
+	// Defaults derived from the parent and chain config. Override
+	// application below layers caller-supplied values on top of these.
+	h := &ethtypes.Header{
+		ParentHash:  parent.Hash(),
```
tomaszslabon, line 143: When I called `eth_simulateV1` with block `4` (`0x4`) on localnet: 
```
  --data '{
    "jsonrpc":"2.0",
    "id":2,
    "method":"eth_simulateV1",
    "params":[
      {
        "blockStateCalls":[
          {
            "calls":[
              {
                "from":"0x1111111111111111111111111111111111111111",
                "to":"0x1111111111111111111111111111111111111111"
              }
            ]
          }
        ]
      },
      "0x4"
    ]
  }' | jq -r '.result[0] | {number,parentHash}'
```
the returned parent hash is:
```
{
  "number": "0x5",
  "parentHash": "0x9954e67e17cde7c6a91e176c38f27d0307365d07331b2a4cf0b49323a8ac404c"
}
```

which is different than the `4`-th block's hash:
```
  --data '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["0x4",false]}' \
  | jq -r '.result.hash'
```
which is:
```
0x65fdad50586258b80fdeec1e9d108e975d20a1a34ab3dfadd97eeedffa0727cc
```

Shouldn't the returned parent hash be the same hash as the block's provided in the command?

lukasz-zimnoch: Good catch — the synthetic header built by `baseHeaderFromContext` only populates the fields the simulate driver consumes, so calling `parent.Hash()` on it produced a value unrelated to the canonical chain hash that `eth_getBlockByNumber` surfaces.

Fixed in 9a70c56a. The rpc backend now derives the canonical Eth-formatted base block hash from the already-fetched Tendermint block via the same `EthHeaderFromTendermint` path `RPCBlockFromTendermintBlock` takes, and forwards it on a new `BaseBlockHash` request field so the keeper can use it as the first simulated block's `parentHash`. Both surfaces now agree.

New coverage: keeper-level `TestSimulateV1_BaseBlockHashOverridesParentHash` pins the override invariant, and rpc-backend-level `TestSimulateV1_PopulatesBaseBlockHash` asserts the canonical hash is forwarded on the gRPC request.

tomaszslabon: The error still seems to be there:
```
  --data '{                                                                                 
    "jsonrpc":"2.0",    
    "id":2,
    "method":"eth_simulateV1",
    "params":[
      {
        "blockStateCalls":[
          {
            "calls":[
              {
                "from":"0x1111111111111111111111111111111111111111",
                "to":"0x1111111111111111111111111111111111111111"
              }
            ]
          }
        ]
      },
      "0x4"
    ]
  }' | jq -r '.result[0] | {number,parentHash}'
 ```
```
{
  "number": "0x5",
  "parentHash": "0x7ab35aebade2c36f1a8c6c14b8003cfdc5a721d359076e38c400d2d16922f532"
}
```
```
  --data '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["0x4",false]}' \
  | jq -r '.result.hash'
```
```
0x2379dc7b36adfcd8408bfbc05b517ab86be6cbf4225ffa746ee2023e0b1a224f
```

File: rpc/backend/simulate_v1.go [hunk 2]
```diff
@@ -0,0 +1,101 @@
...
+	// surfaces for the base block. nil on failure: the keeper keeps its
+	// synthetic-header fallback.
+	var baseBlockHash []byte
+	if blockRes, brErr := b.TendermintBlockResultByNumber(&header.Block.Height); brErr == nil {
+		bloom, _ := b.BlockBloom(blockRes)
+		baseFee, _ := b.BaseFee(blockRes)
+		baseBlockHash = rpctypes.EthHeaderFromTendermint(header.Block.Header, bloom, baseFee).Hash().Bytes()
+	} else {
+		b.logger.Error("eth_simulateV1: base block results not found", "height", header.Block.Height, "error", brErr)
+	}
```
tomaszslabon: Maybe we could simplify it to:
```suggestion
baseBlockHash := common.BytesToHash(header.Block.Header.Hash()).Bytes()
```
This should also help with the [mismatched parent hash problem](https://github.com/mezo-org/mezod/pull/662#discussion_r3138924947).

x/evm/keeper/simulate_v1.go
tomaszslabon: @lukasz-zimnoch 
[This](https://github.com/mezo-org/mezod/pull/662#discussion_r3146453146) could help.

lukasz-zimnoch: The 9a70c56a "fix" was wrong: it forwarded `EthHeaderFromTendermint(...).Hash()`, which is an Ethereum-RLP hash over an *Eth-formatted* shadow header. `eth_getBlockByNumber` on the other side returns `tmtypes.Header.Hash()` via `FormatBlock` — the CometBFT Merkle hash. Two different schemes, hence the persistent mismatch.

Real fix in 28df3881: the rpc backend now forwards `header.Block.Hash().Bytes()` — the canonical CometBFT hash, exactly the value `eth_getBlockByNumber` reports.

I also looked into whether the keeper could resolve this on its own to avoid the wire field entirely. It can't:

- `ctx.BlockHeader()` in the gRPC query context is the *latest* committed header (cosmos-sdk@v0.50.10 baseapp/abci.go:1233 takes it from `app.checkState`), not the queried height's. So `GetHashFn` Case 1's fallback returns the wrong block's hash.
- `stakingKeeper.GetHistoricalInfo(ctx, h)` looked promising, but at FinalizeBlock the SDK only puts a *truncated* `cmtproto.Header` on `ctx` — `ChainID, Height, Time, ProposerAddress, NextValidatorsHash, AppHash` — and PoA's `TrackHistoricalInfo` persists exactly that header. Hashing it produces a different value than CometBFT's block store has. I empirically confirmed this on localnode mid-investigation: the historical-info path mismatched on every height.

So the rpc layer (with access to CometBFT's block store) is the only place the canonical hash is reachable, and forwarding it via `BaseBlockHash` is the right design. The new commit message and the comment in `rpc/backend/simulate_v1.go` capture this so future readers don't repeat my mistake.

rpc/backend/simulate_v1.go
lukasz-zimnoch: Yes, see https://github.com/mezo-org/mezod/pull/662#discussion_r3146885058

x/evm/keeper/grpc_query.go
```diff
@@ -739,3 +790,23 @@ func getChainID(ctx sdk.Context, chainID int64) (*big.Int, error) {
...
+// the SDK context that the gRPC call was anchored at. The returned
+// header only populates the fields the simulate driver consumes
+// (Number, Time, GasLimit, BaseFee, Difficulty, Coinbase). Multi-block
+// support will swap this for a real block fetch so BLOCKHASH resolution
+// lines up with canonical chain history.
+func baseHeaderFromContext(ctx sdk.Context, cfg *statedb.EVMConfig) *ethtypes.Header {
+	return &ethtypes.Header{
+		Number:     big.NewInt(ctx.BlockHeight()),
+		Time:       uint64(ctx.BlockTime().Unix()), //nolint:gosec
+		GasLimit:   mezotypes.BlockGasLimit(ctx),
```
tomaszslabon: @lukasz-zimnoch 
Gas limit calculations seems to be off (`GasLimit` is returned as `0`) on localnet.
I think we should calculate the block gas limit similarly as we do in [BlockMaxGasFromConsensusParams](https://github.com/mezo-org/mezod/blob/main/rpc/types/utils.go#L96), i.e. from consensus params.
See the attached diff.

[gasLimit.txt](https://github.com/user-attachments/files/27125232/gasLimit.txt)

lukasz-zimnoch: Fixed in 0075bfd1, thanks. Routing through `consensusKeeper.Params` like you suggested — `BlockGasLimit(ctx)` was indeed always returning zero on the query path because `CreateQueryContext` doesn't set up `ConsensusParams` or a `BlockGasMeter`. Added a test that wipes consensus params off the ctx to lock the behavior in.

tomaszslabon (approved):

tomaszslabon:
Tests on locanet seems ok.

PR #664, opened 2026-04-24
Multi-call/block eth_simulateV1

lukasz-zimnoch:

### Introduction

Extends `eth_simulateV1` from the single-call / single-block slice on `main` to arbitrarily many blocks each carrying arbitrarily many calls. State mutations propagate across both call and block boundaries through one shared `*statedb.StateDB` allocated for the whole request — covering the EVM journal and mezo's StateDB-scoped cached cosmos ctx (where custom-precompile writes live) uniformly. `traceTransfers`, `validation=true`, `returnFullTransactions`, the request-wide DoS kill switch, and full spec conformance remain follow-up work.

### Changes

- **Multi-call / multi-block driver** (`x/evm/keeper/simulate_v1.go`). One shared StateDB threads through every call of every block. `FinaliseBetweenCalls` clears per-call ephemerals (logs, refund, transient storage, journal, revisions, precompile counter) at the top of every call. `SetTxContext` swaps per-call `TxHash` / `TxIndex` so each call's logs carry distinct values. Cumulative log indices and `BlockHash` are back-stamped after the block hash seals. Headers are built lazily inside the per-block loop and `GasUsed`-patched before the next iteration reads `parent.Hash()`, so the response envelope's parent chain is coherent by construction.
- **Per-call resolvers**. `resolveSimCallNonce` defaults from the shared StateDB so preceding state overrides and CREATE-induced bumps propagate. `resolveSimCallGas` defaults to the per-block remaining budget (`header.GasLimit - cumGasUsed`); explicit-over-budget and zero-remaining paths emit `-38015` as a per-call `SimError` so prior successes survive in the envelope.
- **`BLOCKHASH` resolver**. `newSimGetHashFn` returns canonical hashes for `height <= base.Number` (delegated to `k.GetHashFn(ctx)`, anchored at base by the rpc/backend layer), simulated-sibling hashes for `height > base.Number` via an O(1) height-indexed map of already-finalized siblings, and zero otherwise. Canonical-range hashes are unforgeable by any `BlockOverrides` because `sanitizeSimChain` rejects `sim[]` entries with `Number <= base.Number`.
- **Fork-boundary sentinel**. A single base-derived `params.Rules` drives header construction for every simulated block (matching what `applyMessageWithConfig` reads from ctx internally). `sameForks` compares base vs. last-sim Rules; spans crossing a fork are rejected via `NewSimForkSpanUnsupported` (`-38026`).
- **Intrinsic-gas surfacing**. The driver narrows on `core.ErrIntrinsicGas` from `applyMessageWithConfig` and routes it to a per-call `-38013` (`NewSimIntrinsicGas`) instead of collapsing the request to gRPC Internal.
- **Anchor handshake** (`rpc/backend/simulate_v1.go` + `x/evm/keeper/grpc_query.go`). The backend resolves `BlockNumberOrHash` to a concrete height before marshaling so the keeper anchor validator compares the same numeric value that `ContextWithHeight` anchors ctx at — `BlockNumber`'s JSON round-trip does not preserve sentinels. `validateSimulateV1Anchor` is defense-in-depth for direct-gRPC callers: a numeric mismatch is rejected with `-32602` instead of silently simulating against the wrong base.
- **StateDB additions** (`x/evm/statedb/statedb.go`). `SetTxContext` mirrors go-ethereum's `(*state.StateDB).SetTxContext`. `FinaliseBetweenCalls` now also clears the journal and revision stack — bounds memory across long requests (per-precompile multistore clones don't accumulate) and disarms a latent panic where `addLogChange.Revert` would slice an already-niled `s.logs`.
- **Spec-coded errors** (`x/evm/types/simulate_v1_errors.go`). New constructors: `NewSimBlockGasLimitReached` (`-38015`), `NewSimIntrinsicGas` (`-38013`), `NewSimForkSpanUnsupported` (`-38026`).
- **Test setup fix** (`x/evm/keeper/setup_test.go`). `SetupAppWithT` now populates `ConsensusParams` on the suite ctx; `NewContextLegacy` does not, and `mezotypes.BlockGasLimit` otherwise reads the zero-value fallback.

Feature remains query-side only: no proto changes to `service Msg`, no genesis or stored-state mutations, no ante handler edits, no `ApplyTransaction` edits.

### Testing

- `make test-unit` — passes. New Go coverage in `x/evm/keeper/simulate_v1_test.go` (`TestResolveSimCallNonce_*`, `TestResolveSimCallGas_*`, `TestNewSimGetHashFn_*`, `TestSameForks_*`, `TestNewSimForkSpanUnsupported`) and in `x/evm/keeper/grpc_query_test.go` (`TestSimulateV1_MultiCall_*`, `TestSimulateV1_MultiBlock_*`, `TestSimulateV1_ExplicitGasBelowIntrinsic`).
- `make proto-all` — clean.
- `make lint` — clean on the changed files.
- Hardhat system tests — `./tests/system/system-tests.sh SimulateV1_MultiCall SimulateV1_MultiBlock` runs green against a fresh `make localnode-bin-start`, including the new custom-precompile cases.

---

### Author's checklist

- [x] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`) — docs update for the method and divergences lands with spec conformance.

### Reviewer's checklist

### DIFF
--- a/x/evm/keeper/simulate_v1.go
+++ b/x/evm/keeper/simulate_v1.go
@@ -475,9 +475,14 @@ func (k *Keeper) processSimBlock(
 		// `(*state.StateDB).SetTxContext(thash, ti)` — it only updates
 		// the StateDB's per-call TxHash / TxIndex while leaving the
 		// pre-set BlockHash and LogIndex alone.
+		//
+		// TxIndex is the call's position in the sealed block, not the
+		// loop index: pre-execution rejections above `continue` without
+		// appending to txHashes, so len(txHashes) is the index this
+		// call will land at if it succeeds.
 		callTxHash := computeSimTxHash(msg)
-		callCfg := statedb.NewTxConfig(common.Hash{}, callTxHash, uint(i), 0) //nolint:gosec
-		sdb.SetTxContext(callCfg.TxHash, int(callCfg.TxIndex))                //nolint:gosec
+		callCfg := statedb.NewTxConfig(common.Hash{}, callTxHash, uint(len(txHashes)), 0) //nolint:gosec
+		sdb.SetTxContext(callCfg.TxHash, int(callCfg.TxIndex))                            //nolint:gosec
 
 		res, _, runErr := k.applyMessageWithConfig(
 			ctx,
--- a/x/evm/keeper/simulate_v1.go
+++ b/x/evm/keeper/simulate_v1.go
@@ -1,8 +1,9 @@
 package keeper
 
 import (
-	"fmt"
+	"errors"
 	"math/big"
+	"reflect"
 	"slices"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -251,6 +252,14 @@ func assembleSimBlock(
 // sanitizes the chain, builds preliminary headers, applies state
 // overrides, executes per-call, and assembles response envelopes.
 //
+// State-propagation invariant: ONE *statedb.StateDB is shared across
+// every call and every block in the request. Both the EVM journal and
+// mezo's StateDB-scoped cached-ctx (where custom precompile Cosmos-side
+// writes live) ride on that single StateDB, so call/block continuity
+// covers both layers uniformly. commit=false keeps the whole thing
+// ephemeral. A fresh StateDB per block would silently drop
+// custom-precompile Cosmos writes between blocks.
+//
 // Error policy: spec-coded failures (sanitize-chain violation,
 // state-override validation, per-call preflight) are returned as
 // *types.SimError through the plain error channel; callers branch with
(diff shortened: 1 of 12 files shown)


tomaszslabon:
@lukasz-zimnoch 
I simulated a transfer transaction on an ERC-20 token:
```
  --data '{"jsonrpc":"2.0","id":1,"method":"eth_simulateV1","params":[{"blockStateCalls":[{"stateOverrides":{"0x28c5ef3763a2f1f23d617D3eA81d87846925B1A7":{"balance":"0x56bc75e2d63100000"}},"calls":[{"from":"0x28c5ef3763a2f1f23d617D3eA81d87846925B1A7","to":"0xeF18645ef97D3FaBa46087Cb08a0A2c6029e6876","data":"0xa9059cbb00000000000000000000000011111111111111111111111111111111111111110000000000000000000000000000000000000000000000000de0b6b3a7640000"}]}]},"latest"]}' | jq
```
and got this result:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": [
    {
      "baseFeePerGas": "0x0",
      "calls": [
        {
          "returnData": "0x0000000000000000000000000000000000000000000000000000000000000001",
          "logs": [
            {
              "address": "0xef18645ef97d3faba46087cb08a0a2c6029e6876",
              "topics": [
                "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",
                "0x00000000000000000000000028c5ef3763a2f1f23d617d3ea81d87846925b1a7",
                "0x0000000000000000000000001111111111111111111111111111111111111111"
              ],
              "data": "0x0000000000000000000000000000000000000000000000000de0b6b3a7640000",
              "blockNumber": "0x3ee",
              "transactionHash": "0x29bbc47f6fa292b625f148b11560cb67f1242a6cb2ff8492c611262e30c4124d",
              "transactionIndex": "0x0",
              "blockHash": "0xd98086df2a93c46d0c03e9f8965c8178442f3bc5c22d12a3de36c04e73ac3b7b",
              "logIndex": "0x0",
              "removed": false
            }
          ],
          "gasUsed": "0x4c4b40",
          "status": "0x1"
        }
      ],
      "difficulty": "0x0",
      "extraData": "0x",
      "gasLimit": "0x989680",
      "gasUsed": "0x4c4b40",
      "hash": "0xd98086df2a93c46d0c03e9f8965c8178442f3bc5c22d12a3de36c04e73ac3b7b",
      "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
      "miner": "0xde4fb0b1c619a43d279d3ab4d8673a98a7bd374c",
      "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
      "nonce": "0x0000000000000000",
      "number": "0x3ee",
      "parentHash": "0x7067cbfdc684dd808a5f2a27f567a22f07b2bd7b22bb5caddec824c9797c067b",
      "receiptsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
      "sha3Uncles": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
      "size": "0x0",
      "stateRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
      "timestamp": "0x69f08083",
      "transactions": [
        "0x29bbc47f6fa292b625f148b11560cb67f1242a6cb2ff8492c611262e30c4124d"
      ],
      "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
      "uncles": []
    }
  ]
}
```
Some fields still look incorrect or suspicious:

- `logsBloom` - suspicious if all zeroes; the call emitted a Transfer log, so the bloom should likely reflect that log.
- `stateRoot` - suspicious if all zeroes; that looks like "not computed" rather than a real post-state root.
- `size` - suspicious if 0x0; that looks like placeholder scaffolding, not a real block size.
- `transactionsRoot` - suspicious if it is still the empty-transactions root while transactions is non-empty.
- `receiptsRoot` - suspicious if it is 0x56e81f17...e363b421; that is Ethereum’s canonical empty receipts root, which normally means "no receipts".

lukasz-zimnoch:
`transactionHash` and `blockHash` were the stale-binary effect — your re-test confirms them correct.

For the rest:

`logsBloom`, `transactionsRoot`, `receiptsRoot`, `size` are real bugs. The driver builds the block envelope as a `map[string]interface{}` and never derives roots/bloom/size from execution output. Phase 11 of the plan (commit 921a201b) now enumerates the fix: synthesize per-call `Receipts` and `Transactions`, derive header fields via `CreateBloom` / `DeriveSha`, replace the map with a real `*types.Block` so `block.Size()` returns the RLP body length. New tests assert each field becomes non-trivial when calls execute; a system test re-runs your ERC-20 scenario end-to-end.

`stateRoot` cannot be populated — mezod's `StateDB` wraps a Cosmos cached multistore and has no MPT to call `IntermediateRoot()` on. Echoing `base.Root` would ignore everything the simulation did. Documented as Mezo divergence no. 4 in the same plan amendment, pinned by a regression test.

x/evm/keeper/simulate_v1.go
```diff
@@ -268,138 +277,421 @@ func (k *Keeper) simulateV1(
...
+		// on every emitted log. BlockHash is left zero in callCfg
+		// because the block hash depends on cumulative GasUsed, which
+		// isn't known until every call in this block has executed; the
+		// post-pass below patches log.BlockHash once that hash is
+		// computed. SetTxContext mirrors geth's
+		// `(*state.StateDB).SetTxContext(thash, ti)` — it only updates
+		// the StateDB's per-call TxHash / TxIndex while leaving the
+		// pre-set BlockHash and LogIndex alone.
+		callTxHash := computeSimTxHash(msg)
+		callCfg := statedb.NewTxConfig(common.Hash{}, callTxHash, uint(i), 0) //nolint:gosec
```
tomaszslabon: @lukasz-zimnoch 
There seems to be some indexing mismatch in this scenario: 
- the first transaction in the block gets reverted before execution (e.g. due to gas limit)
- the second transaction (e.g. a transfer) succeeds.

For example, in this call:
```
  --data '{"jsonrpc":"2.0","id":7,"method":"eth_simulateV1","params":[{"blockStateCalls":[{"blockOverrides":{"gasLimit":"0x4c4b40"},"stateOverrides":{"0x28c5ef3763a2f1f23d617D3eA81d87846925B1A7":{"balance":"0x56bc75e2d63100000"}},"calls":[{"from":"0x28c5ef3763a2f1f23d617D3eA81d87846925B1A7","to":"0x1111111111111111111111111111111111111111","value":"0x1","gas":"0x4c4b41"},{"from":"0x28c5ef3763a2f1f23d617D3eA81d87846925B1A7","to":"0xeF18645ef97D3FaBa46087Cb08a0A2c6029e6876","data":"0xa9059cbb00000000000000000000000011111111111111111111111111111111111111110000000000000000000000000000000000000000000000000de0b6b3a7640000","gas":"0x4c4b40"}]}]},"latest"]}' \
  | jq '{transactions:.result[0].transactions, statuses:[.result[0].calls[0].status,.result[0].calls[1].status], first_error:.result[0].calls[0].error, second_log:.result[0].calls[1].logs[0]}'
```
the first transaction gets reverted (due to exceeding gas limit), but the second transaction succeeds (an ERC-20 transfer).

The output is:
```
{
  "transactions": [
    "0x681b197b1e1f5244bdbd22b4e1380741b741ca4ea0a94bdbd525d862e46441b8"
  ],
  "statuses": [
    "0x0",
    "0x1"
  ],
  "first_error": {
    "code": -38015,
    "message": "simulate: block gas limit reached: requested 5000001 > remaining 5000000"
  },
  "second_log": {
    "address": "0xef18645ef97d3faba46087cb08a0a2c6029e6876",
    "topics": [
      "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",
      "0x00000000000000000000000028c5ef3763a2f1f23d617d3ea81d87846925b1a7",
      "0x0000000000000000000000001111111111111111111111111111111111111111"
    ],
    "data": "0x0000000000000000000000000000000000000000000000000de0b6b3a7640000",
    "blockNumber": "0x695",
    "transactionHash": "0x681b197b1e1f5244bdbd22b4e1380741b741ca4ea0a94bdbd525d862e46441b8",
    "transactionIndex": "0x1",
    "blockHash": "0x55a0232455a92f4c037ea959029a988b3e44306df1806ee23d610d133c9ce6fa",
    "logIndex": "0x0",
    "removed": false
  }
}
```
there is only one transaction in the block (`0x681b197b1e1f5244bdbd22b4e1380741b741ca4ea0a94bdbd525d862e46441b8`), but it's index is `0x1` instead of `0x0` in the `second_log`:
```
"transactionIndex": "0x1"
```

However, in the scenario where the first reverted transaction enters the block (i.e. is reverted due to contract logic), the indexing seems ok.

lukasz-zimnoch: Fixed in 6cebd428. Driver was sourcing TxIndex from the loop counter; the in-block index it should have used is len(txHashes) at the time of the call. Pre-execution rejections (over-budget gas, ToMessage failure) skip the txHashes append, so the loop counter outran the surviving call's true position only when an earlier call dropped pre-EVM — matching your in-EVM-revert caveat. Regression test pins the exact shape (oversized first call + log-emitting second call → block has one tx at index 0, log carries transactionIndex=0x0).

tomaszslabon:
I tested various scenarios and multicall/block simulations seem to work correctly, except for the [already acknowledged issues to be solved in the future](https://github.com/mezo-org/mezod/pull/664#issuecomment-4334355384).

tomaszslabon (approved):

PR #666, opened 2026-04-28
StateDB interface compliance

tomaszslabon:
Closes: https://linear.app/thesis-co/issue/MEZO-4005/statedb-interface-compliance

### Introduction

This PR brings Mezod’s custom `StateDB` in line with the `go-ethereum v1.16.9` interface.

The main change is adding the new `Finalise()` method for interface compliance and introducing `ResetTxEphemerals()` for flows that reuse a `StateDB` between simulated calls, such as `eth_simulateV1`. Normal on-chain transaction execution is unchanged: Mezod still creates and discards a `StateDB` per transaction.

The PR also implements `AccessEvents()` and finishes `GetStateAndCommittedState()`. Mezod does not support Verkle / EIP-4762 today, so `AccessEvents()` returns `nil`.

### Changes

- Implemented `Finalise()` as a no-op interface-compliance stub.
- Added `ResetTxEphemerals()` for simulation call boundaries.
- Implemented `AccessEvents()`.
- Finished `GetStateAndCommittedState()`.
- Added coverage for new `StateDB` return values and reset behavior.

### Testing

- `go test ./x/evm/statedb`

---

### Author's checklist

- [X] Updated relevant unit and integration tests

### Reviewer's checklist

### DIFF
--- a/x/evm/statedb/statedb.go
+++ b/x/evm/statedb/statedb.go
@@ -434,9 +434,33 @@ func (s *StateDB) AccessEvents() *gethstate.AccessEvents {
 	return nil
 }
 
+// Finalise clears per-call state for flows that reuse one StateDB without
+// committing between calls, such as simulations. Normal transaction execution
+// does not call it because Mezod creates and discards a StateDB per transaction.
+//
+// State objects, access list, and the live cached Cosmos context are preserved,
+// so accumulated cross-call account/storage mutations and precompile-side writes
+// remain visible. The journal and revision stack are cleared to bound memory and
+// to prevent reverting across call boundaries. Callers must not hold a live
+// Snapshot() across this call.
+//
 //nolint:misspell
-func (s *StateDB) Finalise(bool) {
-	// TODO (geth-upgrade): implement when adding support for the new state finalization flow in 1.16.9.
+func (s *StateDB) Finalise(deleteEmptyObjects bool) {
+	// Required by geth's StateDB interface. Mezod does not use Finalise for
+	// canonical empty-account deletion.
+	_ = deleteEmptyObjects
+
+	s.logs = nil
+	s.refund = 0
+	s.transientStorage = newTransientStorage()
+	s.ongoingPrecompilesCallsCounter = 0
+	s.journal = newJournal()
+	s.validRevisions = nil
+	s.nextRevisionID = 0
+
+	for _, obj := range s.stateObjects {
+		obj.newContract = false
+	}
 }
 
 // GetCode returns the code of account, nil if not exists.
@@ -821,38 +845,6 @@ func (ccc *CachedCtxCheckpoint) Revert(stateDB *StateDB) {
 		// at time of creation
 		ccc.ms.Write()
 	}
-}
-
-// FinaliseBetweenCalls clears per-call ephemeral state (logs, refund,
-// transient storage, journal, and revision stack) and resets the
-// precompile-call counter while preserving state objects, access list,
-// and the live cached cosmos ctx — so accumulated cross-call
-// account/storage mutations and precompile-side writes remain visible.
-// Used by simulate drivers running several calls against a shared
-// StateDB without committing between them.
-//
-// Clearing the journal bounds memory: each custom-precompile call
-// appends a multistore Clone() to the journal, which would otherwise
-// accumulate across the whole request. It also disarms the latent
-// panic where addLogChange.Revert reads s.logs[len-1] after s.logs has
-// been niled by a prior FinaliseBetweenCalls.
-//
-// Invariant: callers must not hold a live Snapshot() across this call.
-// The revision stack is wiped, so a stale revision id will fail
-// RevertToSnapshot. The simulate driver does not snapshot at the
-// top level; future drivers must preserve that.
-//
-// TODO (geth-upgrade): v1.16.9 introduces the canonical-spelled geth
-// finaliser on the [vm.StateDB] interface; once the upgrade lands,
-// this helper collapses into that call plus the counter reset.
-func (s *StateDB) FinaliseBetweenCalls() {
-	s.logs = nil
-	s.refund = 0
-	s.transientStorage = newTransientStorage()
-	s.ongoingPrecompilesCallsCounter = 0
-	s.journal = newJournal()
-	s.validRevisions = nil
-	s.nextRevisionID = 0
 }
 
 func (s *StateDB) CacheContext() (sdk.Context, *CachedCtxCheckpoint) {
--- a/x/evm/statedb/statedb.go
+++ b/x/evm/statedb/statedb.go
@@ -510,7 +510,6 @@ func (s *StateDB) GetCommittedState(addr common.Address, hash common.Hash) commo
 
 // GetStateAndCommittedState returns the current value and the committed value.
 func (s *StateDB) GetStateAndCommittedState(addr common.Address, hash common.Hash) (common.Hash, common.Hash) {
-	// TODO (geth-upgrade): check when reviewing state access changes in 1.16.9.
 	stateObject := s.getStateObject(addr)
 	if stateObject != nil {
 		return stateObject.GetState(hash), stateObject.GetCommittedState(hash)
--- a/x/evm/statedb/statedb.go
+++ b/x/evm/statedb/statedb.go
@@ -450,7 +450,6 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) {
 		panic("unsupported on Mezo: empty-account deletion during Finalise")
 	}
 
-	s.logs = nil
 	s.refund = 0
 	s.transientStorage = newTransientStorage()
 	s.ongoingPrecompilesCallsCounter = 0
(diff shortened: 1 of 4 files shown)


x/evm/statedb/statedb.go
```diff
@@ -430,13 +430,39 @@ func (s *StateDB) Witness() *stateless.Witness {
...
+	s.refund = 0
+	s.transientStorage = newTransientStorage()
+	s.ongoingPrecompilesCallsCounter = 0
+	s.journal = newJournal()
+	s.validRevisions = nil
+	s.nextRevisionID = 0
+
+	for _, obj := range s.stateObjects {
+		obj.newContract = false
+	}
```
lukasz-zimnoch, line 491: Can we document why we need to perform this loop? See https://github.com/mezo-org/mezod/pull/666#discussion_r3200750815 where I proposed such a clarification.

x/evm/statedb/statedb.go
```diff
@@ -430,13 +430,39 @@ func (s *StateDB) Witness() *stateless.Witness {
...
+// State objects, access list, and the live cached Cosmos context are preserved,
+// so accumulated cross-call account/storage mutations and precompile-side writes
+// remain visible. The journal and revision stack are cleared to bound memory and
+// to prevent reverting across call boundaries. Callers must not hold a live
+// Snapshot() across this call.
+//
 //nolint:misspell
-func (s *StateDB) Finalise(bool) {
-	// TODO (geth-upgrade): implement when adding support for the new state finalization flow in 1.16.9.
+func (s *StateDB) Finalise(deleteEmptyObjects bool) {
```
lukasz-zimnoch: After thinking more about it, it seems that we should split this into a no-op `Finalise(bool)` for interface compliance and a separately-named method that does the actual cleanup. The initial idea for using `Finalise` for sim calls was probably not the best one.

Right now `Finalise(false)` at `simulate_v1.go:516` reads oddly. Geth's `Finalise` flushes dirty objects toward a trie root and may delete EIP-158-empty accounts, and we don't do either — we use a fresh `StateDB` per on-chain tx and account-level deletion lives at the keeper layer. So the name doesn't carry useful information at the call site; you have to look up the geth contract and then notice ours diverges.

No geth code path actually reaches `Finalise` on our `StateDB` — the only caller is the simulate driver. So `Finalise(bool)` only exists for interface compliance and can be a stub. Former `FinaliseBetweenCalls` could be renamed to `ResetTxEphemerals` to better reflect the intention:

```go
  // Finalise satisfies vm.StateDB on go-ethereum v1.16.9. It's a no-op
  // stub kept solely for interface compliance.
  //
  // Geth's Finalise flushes dirty objects toward a trie root and may
  // delete EIP-158-empty accounts. Mezo does neither, because it uses
  // a fresh StateDB per tx and account-level deletion lives at the
  // keeper layer.
  //
  //nolint:misspell
  func (s *StateDB) Finalise(_ bool) {}

  // ResetTxEphemerals clears per-call ephemeral state (logs, refund,
  // transient storage, journal, revision stack, and the EIP-6780
  // newContract flag) and resets the precompile-call counter while
  // preserving state objects, access list, and the live cached cosmos
  // ctx — so accumulated cross-call account/storage mutations and
  // precompile-side writes remain visible. Used by simulate drivers
  // running several calls against a shared StateDB without committing
  // between them.
  //
  // Clearing the journal bounds memory: each custom-precompile call
  // appends a multistore Clone() to the journal, which would otherwise
  // accumulate across the whole request. It also disarms the latent
  // panic where addLogChange.Revert reads s.logs[len-1] after s.logs
  // has been niled by a prior ResetTxEphemerals call.
  //
  // Invariant: callers must not hold a live Snapshot() across this call.
  // The revision stack is wiped, so a stale revision id will fail
  // RevertToSnapshot. The simulate driver does not snapshot at the top
  // level; future drivers must preserve that.
  func (s *StateDB) ResetTxEphemerals() {
        s.logs = nil
        s.refund = 0
        s.transientStorage = newTransientStorage()
        s.ongoingPrecompilesCallsCounter = 0
        s.journal = newJournal()
        s.validRevisions = nil
        s.nextRevisionID = 0

        // EIP-6780: each simulated call is a separate transaction, so
        // newContract has to be cleared between calls — otherwise a
        // contract created in call N could be 6780-destroyed in call N+1.
        // Geth handles this at end-of-tx by iterating journal.dirties;
        // we just reset the journal above, so dirties is empty by the
        // time this loop runs. Iterating live state objects sidesteps
        // that ordering dependency, and the assignment is a no-op for
        // anything that never had the flag.
        for _, obj := range s.stateObjects {
                obj.newContract = false
        }
  }
```
`simulate_v1.go:516` then calls `sdb.ResetTxEphemerals()`.

x/evm/statedb/statedb.go
tomaszslabon: Done in 2f793ad84a42216ad60edd82b3b5e0a964b5c98b.

x/evm/statedb/statedb.go
tomaszslabon: Done in 2f793ad84a42216ad60edd82b3b5e0a964b5c98b.

x/evm/statedb/statedb.go
```diff
@@ -430,13 +430,60 @@ func (s *StateDB) Witness() *stateless.Witness {
...
+// precompile-side writes remain visible. Used by simulate drivers
+// running several calls against a shared StateDB without committing
+// between them.
+//
+// Clearing the journal bounds memory: each custom-precompile call
+// appends a multistore Clone() to the journal, which would otherwise
+// accumulate across the whole request. It also disarms the latent
+// panic where addLogChange.Revert reads s.logs[len-1] after s.logs
+// has been cleared by a prior ResetTxEphemerals call.
+//
```
lukasz-zimnoch, line 462: After this runs, self-destructed objects stay live in `s.stateObjects` — no `deleted`-flag sweep like geth's `Finalise` does. So in a simulate sequence, call N+1 still sees `Exist(addr)==true` for an account destructed in call N. Pre-existing, RPC-only impact, fine to keep — but worth a docstring note so simulate clients aren't surprised.

x/evm/statedb/statedb.go
```diff
@@ -430,13 +430,60 @@ func (s *StateDB) Witness() *stateless.Witness {
...
+// ctx — so accumulated cross-call account/storage mutations and
+// precompile-side writes remain visible. Used by simulate drivers
+// running several calls against a shared StateDB without committing
+// between them.
+//
+// Clearing the journal bounds memory: each custom-precompile call
+// appends a multistore Clone() to the journal, which would otherwise
+// accumulate across the whole request. It also disarms the latent
+// panic where addLogChange.Revert reads s.logs[len-1] after s.logs
+// has been cleared by a prior ResetTxEphemerals call.
```
lukasz-zimnoch, line 461: The loop is only exercised with a single state object today. A regression ranging the wrong map or skipping entries wouldn't surface. Worth extending the test: `CreateContract` on two addresses, run the reset, assert `SelfDestruct6780` returns `destroyed=false` for both.

x/evm/statedb/statedb.go
```diff
@@ -430,13 +430,60 @@ func (s *StateDB) Witness() *stateless.Witness {
 }
 
 func (s *StateDB) AccessEvents() *gethstate.AccessEvents {
-	// TODO (geth-upgrade): implement when adding support for the new access-event flow in 1.16.9.
+	// Mezo does not activate Verkle/EIP-4762 and does not rely on Verkle trees.
 	return nil
```
lukasz-zimnoch, line 434: No test pins this returning nil. A future refactor that swaps it for a non-nil zero value would slip past CI. One-liner would do: `suite.Require().Nil(db.AccessEvents())`.

x/evm/statedb/statedb.go
```diff
@@ -430,13 +430,60 @@ func (s *StateDB) Witness() *stateless.Witness {
...
+func (s *StateDB) Finalise(_ bool) {}
+
+// ResetTxEphemerals clears per-call ephemeral state (logs, refund,
+// transient storage, journal, revision stack, and the EIP-6780
+// newContract flag) and resets the precompile-call counter while
+// preserving state objects, access list, and the live cached cosmos
+// ctx — so accumulated cross-call account/storage mutations and
+// precompile-side writes remain visible. Used by simulate drivers
+// running several calls against a shared StateDB without committing
+// between them.
```
lukasz-zimnoch, line 455: The counter reset isn't directly asserted — only implied by `TestFinaliseBetweenCalls`'s "cap is not exceeded" comment. Dropping this line wouldn't fail anything. Worth adding a test that registers `maxPrecompilesCallsPerExecution+1` checkpoints across the reset; the (n+1)-th only succeeds if the counter was zeroed.

x/evm/statedb/statedb.go
tomaszslabon: Done in 16bce68bfdf739e210f9e495fdd81e18e8d1e65a.

x/evm/statedb/statedb.go
tomaszslabon: Done in 16bce68bfdf739e210f9e495fdd81e18e8d1e65a.

x/evm/statedb/statedb.go
tomaszslabon: Done in 16bce68bfdf739e210f9e495fdd81e18e8d1e65a.

x/evm/statedb/statedb.go
tomaszslabon: Done in 16bce68bfdf739e210f9e495fdd81e18e8d1e65a.

lukasz-zimnoch (approved):

PR #665, opened 2026-04-27
DoS hardening for eth_simulateV1

lukasz-zimnoch:
References: [MEZO-4227](https://linear.app/thesis-co/issue/MEZO-4227)

### Introduction

Hardens `eth_simulateV1` against denial-of-service. Operators get a single config flag that hides the JSON-RPC method behind a `-32601`. The request envelope is bounded on both block count and cumulative call count at the keeper input boundary — once, ahead of any allocation. The request-wide gas pool is capped through `json-rpc.gas-cap`, and a per-request deadline backed by an EVM cancel watcher prevents one client from holding a goroutine indefinitely. RPC-config-dependent bounds (kill switch, gas-cap, timeout) live in the RPC layer; RPC-config-independent bounds (block / call counts) live in the keeper. Operators who need to suppress simulate at the gRPC layer must additionally restrict the SDK gRPC port — that constraint is documented in the toml template, on `Keeper.SimulateV1`, and on the relevant constructors.

### Changes

- **Operator kill switch** (`server/config/config.go`, `server/config/toml.go`, `rpc/backend/backend.go`, `rpc/backend/node_info.go`, `rpc/namespaces/ethereum/eth/simulate_v1.go`). A new `json-rpc.simulate-disabled` app.toml flag, surfaced through `EVMBackend.SimulateDisabled()`, makes `PublicAPI.SimulateV1` short-circuit before the backend with a `-32601` `SimError` shaped to match the JSON-RPC framework's "method not registered" message. A node hiding the endpoint is indistinguishable from one that does not implement it. The toml comment notes that the flag covers the JSON-RPC namespace only — operators who need to block simulate at the SDK gRPC port must firewall it separately.
[...]

PR #667, opened 2026-04-28
Final Geth parity for eth_simulateV1

lukasz-zimnoch:

### Introduction

Closes the remaining feature-parity gap between mezod's `eth_simulateV1` and the execution-apis spec by wiring three optional behaviours that the DoS-hardened driver from #665 does not yet provide: `traceTransfers` (synthetic ERC-7528 Transfer logs), `validation=true` (tx-level preflight failures as top-level fatals), and `returnFullTransactions` plus a real `*ethtypes.Block` envelope (so `logsBloom`, `transactionsRoot`, `receiptsRoot`, and `size` derive correctly). Behaviour with the flags off is unchanged.

### Changes

- **`traceTransfers` synthetic logs** (`x/evm/keeper/simulate_v1_logtracer.go`, `x/evm/keeper/simulate_v1.go`, `x/evm/statedb/statedb.go`). A new unexported `simTracer` in the keeper package (mirroring upstream go-ethereum's `internal/ethapi/logtracer.go`, which co-locates the analogous helper next to `simulate.go`) implements a per-frame log stack: `OnEnter` pushes a frame and emits a synthetic ERC-7528 `Transfer(address,address,uint256)` log at `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` for value-bearing CALL / CALLCODE / CREATE / CREATE2 / SELFDESTRUCT edges; DELEGATECALL is excluded because the value belongs to the caller's frame. Real EVM logs flow through `OnLog`, captured in the current frame. On revert the frame's logs are dropped, on success they merge into the parent. The tracer reuses the keeper's `mezoCustomPrecompileAddrs` set (built from `types.DefaultPrecompilesVersions`) so the synthetic emission never double-counts an event the precompile itself emitted (e.g. `btctoken.transfer`). `processSimBlock` constructs a fresh tracer per simulated block, wires it via `vm.Config.Tracer` and `statedb.SetTracingHooks`, and substitutes the tracer's logs into each `SimCallResult` so they ride the existing post-block `BlockHash`/`Index` back-stamp loop. `StateDB.SetTracingHooks` is the single new seam on the StateDB — it gives the tracer's `OnLog` hook the same `*tracing.Hooks` value already used by the EVM's call hooks; the consensus path never sets hooks so the dispatch is a no-op.
- **`validation=true` pre-call gate** (`x/evm/keeper/simulate_v1.go`, `x/evm/types/simulate_v1_errors.go`). Six new `NewSim*` constructors cover the spec-reserved JSON-RPC codes that flow through the request-level fatal channel under validation: `-38010` (nonce too low), `-38011` (nonce too high), `-38012` (block base-fee override below the chain-computed floor), `-38014` (insufficient funds), `-38025` (init-code over `params.MaxInitCodeSize`), and `-32005` (tx fee-cap below block base-fee). The new `validateSimCall` helper runs the per-call gates in geth's actual state-transition order — nonce → fee-cap-vs-baseFee → balance → init-code → intrinsic gas — and returns a typed `*SimError` that the existing `simulateV1ErrResponse` routing surfaces as a top-level fatal. The `-38012` block check stays in `processSimBlock` because it depends on the parent header and the `BlockOverrides.BaseFeePerGas` pointer, not on the per-call message. `msg.SkipAccountChecks = true` regardless of validation flag (state overrides may make `from` a contract). Revert and VM errors stay per-call; the gate is bypassed entirely when `opts.Validation = false`.
- **`returnFullTransactions` and full block envelope** (`x/evm/types/simulate_v1.go`, `x/evm/keeper/simulate_v1.go`). `processSimBlock` materialises a per-call `*ethtypes.Transaction` and `*ethtypes.Receipt` (Type / Status / CumulativeGasUsed / Logs / Bloom / TxHash / GasUsed / BlockNumber / TransactionIndex / ContractAddress on CREATE), then assembles a real `*ethtypes.Block` via `ethtypes.NewBlock`, which derives `header.TxHash`, `header.ReceiptHash`, and `header.Bloom` from the supplied lists. The synthetic transaction envelope is selected per request fields, mirroring go-ethereum v1.16's `internal/ethapi.TransactionArgs.ToTransaction(DynamicFeeTxType)`: callers who set `gasPrice` get a `LegacyTx` (the legacy fee field force-overrides the default type), every other shape gets a `DynamicFeeTx` with `accessList` nested when present. ChainID flows in from the chain config so type-1/2 hashes match the network. The driver hands the assembled block, the per-block `Senders []common.Address` (keyed by call index — keying by hash would collide when distinct callers issue identical args, since the synthetic tx has no signer), the `fullTx` flag, the chain config, and the per-call results to `types.NewSimBlockResult`, which runs an unexported `marshalSimBlock` helper to render the spec-shaped envelope map up front. The wire shape of `SimBlockResult` is therefore a thin `Block map[string]interface{}` plus `Calls []SimCallResult`; `MarshalJSON` flattens the two with no branching, and `UnmarshalJSON` rebuilds the same `Block` map from the gRPC payload bytes. The block / header marshaling helpers (`marshalEthHeader`, `marshalEthBlock`, `rpcTransaction`, `newRPCTransaction*`) are ported from go-ethereum's `internal/ethapi/api.go` (LGPL header preserved) and live in `x/evm/types/simulate_v1.go` as unexported symbols since nothing outside this file consumes them — placement avoids the import cycle with `rpc/types`.
- **Mezo divergence pinned** (`x/evm/keeper/grpc_query_test.go`, `x/evm/types/simulate_v1_test.go`). `header.Root` stays the zero hash (mezod's `statedb.StateDB` wraps a Cosmos cached multistore and has no MPT to call `IntermediateRoot()` on); echoing `base.Root` would be misleading. Tests pin `stateRoot == 0x0…0` at both the type-marshal and keeper-E2E layers so an accidental future change surfaces loudly.

### Testing

- `make test-unit` — green except `TestSimulateV1_Timeout_MidCall`, a pre-existing wall-clock-sensitive flake on this VM (the same test fails on the parent branch tip; not introduced here).
- New Go coverage:
  - `x/evm/keeper/simulate_v1_logtracer_test.go`: 14 unit tests covering the happy path, every opcode the tracer cares about, deny-list, nested-revert log drops, and per-frame context stamping.
  - `x/evm/keeper/grpc_query_test.go` extensions: 12 keeper E2E tests for `traceTransfers`; 14 for `validation=true` (one per fatal code plus revert-stays-per-call, NoBaseFee-ignored, abort-aware multi-block, determinism); 11 `validation=false` twins; 17 `BlockEnvelope` and `FullTx` cases including independent `DeriveSha` / `Bloom` recomputation, multi-sender match-by-hash, hash-stable, the empty-block-uses-empty-roots gap-fill case, and envelope-shape pins for the gasPrice-forces-legacy override and the dynamic-fee shape with a nested access list.
  - `x/evm/keeper/simulate_v1_test.go` extensions: 14 `validateSimCall` white-box tests covering each gate's pass/fail edge, fork-gating, ordering, and determinism.
  - `x/evm/types/simulate_v1_test.go` extensions: 12 `SimBlockResult.MarshalJSON` cases — full envelope (hash-only and FullTx with sender patch), multi-tx-two-senders match by call index, empty-block roots, defensive nil/missing senders, `stateRoot=0` pin, `size > 0`, `logsBloom` cross-check, plus a gRPC-bytes round-trip that confirms the keeper-built envelope survives JSON-marshal and unmarshal back into a fresh `SimBlockResult`.
  - `x/evm/types/simulate_v1_errors_test.go` extensions: one happy-path test per new constructor + boundary + an `errors.As` round-trip over all of them.
  - `rpc/backend/simulate_v1_test.go` extensions: full-envelope passthrough through gRPC `Result`, opts-blob preservation for `returnFullTransactions`.
- New system tests (Hardhat): `SimulateV1_TraceTransfers.test.ts`, `SimulateV1_Validation.test.ts`, `SimulateV1_BlockEnvelope.test.ts`, `SimulateV1_FullTx.test.ts`. Not run in CI here (no localnode in this sandbox); the .ts files match the existing `SimulateV1_*.test.ts` style and run via `tests/system/system-tests.sh`.
- `make lint` — green on the changed packages (`x/evm/{types,keeper,statedb}`, `rpc/backend`).

---

### Author's checklist

- [x] Updated the documentation
- [x] Updated the changelog


x/evm/keeper/simulate_v1.go
```diff
@@ -749,18 +821,72 @@ func sameForks(a, b params.Rules) bool {
...
+// buildSimTx materializes the synthetic *ethtypes.Transaction for a
+// simulated call. Returning the transaction (rather than just the hash)
+// lets the driver pass the same value into NewBlock to derive
+// transactionsRoot and into the Senders map without rebuilding it.
+//
+// Simulate txs are unsigned legacy txs: V/R/S = 0, signing fields left
+// zero. tx.Hash() is therefore stable and distinct per (nonce, to,
+// value, data, gas, gasPrice) tuple — sufficient for the assembled
+// block's per-call identifiers.
+func buildSimTx(msg core.Message) *ethtypes.Transaction {
```
tomaszslabon: Type access-list transaction simulation request looks like this:
 ```
  --data '{
    "jsonrpc":"2.0",
    "id":33,
    "method":"eth_simulateV1",
    "params":[
      {
        "returnFullTransactions": true,
        "blockStateCalls":[
          {
            "calls":[
              {
                "from":"0x2D3eAf81DaBcE3C7B6CcF2BfBE9FE24053e3354c",
                "to":"0xef8607235F168C6044ce0eB730F428a289ECdd14",
                "value":"0x1",
                "accessList":[]
              }
            ]
          }
        ]
      },
      "latest"
    ]
  }'
```
It is type `1` because:
- `accessList` is provided
- `gasPrice`  and `maxFeePerGas` are not present (which would change the type to `0` or `2`).

The response is:
```
{"jsonrpc":"2.0","id":33,"result":[{"baseFeePerGas":"0x0","calls":[{"returnData":"0x","logs":[],"gasUsed":"0x4c4b40","status":"0x1"}],"difficulty":"0x0","extraData":"0x","gasLimit":"0x989680","gasUsed":"0x4c4b40","hash":"0x963c1dabda36af2ca7825a205b655ab6e9c2300cd870f7e2cbcba2776da34d88","logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","miner":"0xd0bf29bc03d3fd1b5cabd2f829e46fc48982f89d","mixHash":"0x0000000000000000000000000000000000000000000000000000000000000000","nonce":"0x0000000000000000","number":"0x899","parentHash":"0x5d97ca61376d105cec9e18bd849a45dcdd377f0b26d960f4684aa655bb2cc89e","receiptsRoot":"0x211ba78f9407c2e3f6fd84346af74ec38f6bd80f33435bf0f75594635b9fba0b","sha3Uncles":"0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347","size":"0x223","stateRoot":"0x0000000000000000000000000000000000000000000000000000000000000000","timestamp":"0x69f222ac","transactions":[{"blockHash":"0x963c1dabda36af2ca7825a205b655ab6e9c2300cd870f7e2cbcba2776da34d88","blockNumber":"0x899","chainId":"0x7b7b","from":"0x2d3eaf81dabce3c7b6ccf2bfbe9fe24053e3354c","gas":"0x989680","gasPrice":"0x0","hash":"0xc47fe90d82a2773256fc7004a5ad0a0a846f5edcfe7ef65b29ca17ef3160c532","input":"0x","nonce":"0x3","r":"0x0","s":"0x0","to":"0xef8607235f168c6044ce0eb730f428a289ecdd14","transactionIndex":"0x0","type":"0x0","v":"0x0","value":"0x1"}],"transactionsRoot":"0xa8eb40af400b95d921adc6bb1386fb82047328e9df0bc3b3e53a5b761ec91663","uncles":[]}]}
```
The incorrect fields seem to be:
- `transactions[0].type` = "0x0" - this request is access-list-shaped, and on Ethereum eth_simulateV1 the returned full tx is typed, not legacy
- `transactions[0].accessList` - missing entirely; it should be present as []
- `transactions[0].gasPrice` = "0x0" - suspicious side effect of flattening the tx into legacy form
- `transactions[0].hash` - likely wrong as a consequence of using the wrong tx type/encoding
- `transactionsRoot` - likely wrong for the same reason
- top-level block `hash` - likely wrong as a consequence, because it includes `transactionsRoot`

File: x/evm/keeper/simulate_v1.go [hunk 2]
```diff
@@ -749,18 +821,72 @@ func sameForks(a, b params.Rules) bool {
...
+	return ethtypes.NewTx(&ethtypes.LegacyTx{
```
tomaszslabon: A dynamic fee transaction simulation looks like this:
```
  --data '{
    "jsonrpc":"2.0",
    "id":35,
    "method":"eth_simulateV1",
    "params":[
      {
        "returnFullTransactions": true,
        "blockStateCalls":[
          {
            "calls":[
              {
                "from":"0x2D3eAf81DaBcE3C7B6CcF2BfBE9FE24053e3354c",
                "to":"0xef8607235F168C6044ce0eB730F428a289ECdd14",
                "value":"0x1",
                "maxFeePerGas":"0x64",
                "maxPriorityFeePerGas":"0x2"
              }
            ]
          }
        ]
      },
      "latest"
    ]
  }'
```
The response is:
```
{"jsonrpc":"2.0","id":35,"result":[{"baseFeePerGas":"0x0","calls":[{"returnData":"0x","logs":[],"gasUsed":"0x4c4b40","status":"0x1"}],"difficulty":"0x0","extraData":"0x","gasLimit":"0x989680","gasUsed":"0x4c4b40","hash":"0x6553b7f1bb8d7a440ff784fb70f9e4324520669510b423397d17567415362427","logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","miner":"0x756390af6fa308405dd2990d954c65a54b040bdc","mixHash":"0x0000000000000000000000000000000000000000000000000000000000000000","nonce":"0x0000000000000000","number":"0x922","parentHash":"0x5b26d45c544c21aeb6d14271a7dc4ae303a0a3c4ba1a2a4d2d69fbf754088c9d","receiptsRoot":"0x211ba78f9407c2e3f6fd84346af74ec38f6bd80f33435bf0f75594635b9fba0b","sha3Uncles":"0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347","size":"0x223","stateRoot":"0x0000000000000000000000000000000000000000000000000000000000000000","timestamp":"0x69f22491","transactions":[{"blockHash":"0x6553b7f1bb8d7a440ff784fb70f9e4324520669510b423397d17567415362427","blockNumber":"0x922","chainId":"0x7b7b","from":"0x2d3eaf81dabce3c7b6ccf2bfbe9fe24053e3354c","gas":"0x989680","gasPrice":"0x2","hash":"0x7a5384a9375832c0b5ca20c413cb8d447f108b1daedcb6eddc52773e07a6a283","input":"0x","nonce":"0x3","r":"0x0","s":"0x0","to":"0xef8607235f168c6044ce0eb730f428a289ecdd14","transactionIndex":"0x0","type":"0x0","v":"0x0","value":"0x1"}],"transactionsRoot":"0x46279fb97674416defc9fa22e87c37dfa666f735fd93cfdb0b3748a136127aba","uncles":[]}]}
```

What looks incorrect:
- `transactions[0].type` = "0x0" - wrong; this request is dynamic-fee-shaped, so the returned full tx should be typed, i.e. 0x2
- `transactions[0].maxFeePerGas` - missing
- `transactions[0].maxPriorityFeePerGas` - missing
- `transactions[0].gasPrice` = "0x2" - suspicious side effect of flattening the tx into legacy form
- `transactions[0].hash` - likely wrong because it is derived from the wrong tx type/encoding
- `transactionsRoot` - likely wrong for the same reason
- top-level block `hash` - likely wrong as a consequence, because it includes transactionsRoot

x/evm/keeper/simulate_v1.go
lukasz-zimnoch: Fixed in 12cfa70b. `buildSimTx` now selects the envelope per request fields, mirroring geth v1.16's `internal/ethapi.TransactionArgs.ToTransaction(DynamicFeeTxType)`: callers who set `gasPrice` get a `LegacyTx`, everything else gets a `DynamicFeeTx` with `accessList` nested. So this access-list-shaped request now returns type `0x2` rather than `0x1` — under v1.16's simulator the access-list case in the `ToTransaction` switch is unreachable because `defaultType==DynamicFeeTxType` always wins first when no explicit fee fields are set. The `accessList`, `chainId`, `maxFeePerGas`, `maxPriorityFeePerGas`, hash, transactionsRoot, and block hash are now all correct for the typed envelope.

x/evm/keeper/simulate_v1.go
lukasz-zimnoch: Fixed in 12cfa70b. The dynamic-fee shape now returns a real `DynamicFeeTx` envelope: `type=0x2`, `chainId`, `maxFeePerGas`, `maxPriorityFeePerGas`, `accessList`, with the tx hash / transactionsRoot / block hash all derived from the typed encoding.

x/evm/keeper/simulate_v1.go
```diff
@@ -592,18 +624,59 @@ func (k *Keeper) processSimBlock(
...
+
+	// Back-stamp the values that aren't knowable until the block is
+	// sealed: log.BlockHash (depends on cumulative GasUsed via
+	// header.Hash()), log.Index (must be per-block monotonic; AddLog
+	// stamps from txConfig.LogIndex+len(s.logs), both of which reset
+	// between calls), and receipt.BlockHash (mirrors the per-log stamp
+	// for envelope consistency).
 	var logIdx uint
 	for i := range calls {
 		for _, log := range calls[i].Logs {
```
tomaszslabon, line 687: @lukasz-zimnoch I executed the following scenario: in one block:
1) a BTC precompile `transfer(...)`
2) a regular ERC-20 `transfer(...)`
```
  --data '{
    "jsonrpc":"2.0",
    "id":411,
    "method":"eth_simulateV1",
    "params":[
      {
        "returnFullTransactions": true,
        "blockStateCalls":[
          {
            "calls":[
              {
                "from":"0x2D3eAf81DaBcE3C7B6CcF2BfBE9FE24053e3354c",
                "to":"0x7b7c000000000000000000000000000000000000",
                "data":"0xa9059cbb000000000000000000000000111111111111111111111111111111111111111100000000000000000000000000000000000000000000000006f05b59d3b20000"
              },
              {
                "from":"0x2D3eAf81DaBcE3C7B6CcF2BfBE9FE24053e3354c",
                "to":"0x745D6B0b19bA4746EDE3125E61811A1F89E6BB58",
                "data":"0xa9059cbb00000000000000000000000011111111111111111111111111111111111111110000000000000000000000000000000000000000000000000de0b6b3a7640000"
              }
            ]
          }
        ]
      },
      "latest"
    ]
  }' | jq '{blockNumber:.result[0].number, btcLogBlockNumber:.result[0].calls[0].logs[0].blockNumber, erc20LogBlockNumber:.result[0].calls[1].logs[0].blockNumber, blockHash:.result[0].hash, btcLogBlockHash:.result[0].calls[0].logs[0].blockHash}'
```

The result is:
```
{
  "blockNumber": "0xe9c",
  "btcLogBlockNumber": "0xe9b",
  "erc20LogBlockNumber": "0xe9c",
  "blockHash": "0x9018cd85cd3e8a02adc5c4d7f7e95069349f608e7ac060bd7bee11831a86f791",
  "btcLogBlockHash": "0x9018cd85cd3e8a02adc5c4d7f7e95069349f608e7ac060bd7bee11831a86f791"
}
```
It shows a mismatch between:
`.result[0].number`: "0xe9c",
`.result[0].calls[0].logs[0].blockNumber`: "0xe9b"

`.result[0].calls[0].logs[0].blockNumber` is showing the incorrect block number. It should show "0xe9c".

The fix could be changing:
```
	for i := range calls {
		for _, log := range calls[i].Logs {
			log.BlockHash = finalBlockHash
			log.Index = logIdx
			logIdx++
		}
	}
```
to:
```
	for i := range calls {
		for _, log := range calls[i].Logs {
			log.BlockHash = finalBlockHash
			log.BlockNumber = header.Number.Uint64()
			log.Index = logIdx
			logIdx++
		}
	}
```

lukasz-zimnoch: Fixed in 4cbd020e — the back-stamp loop now also pins log.BlockNumber to header.Number alongside the existing BlockHash/Index pass. Added a keeper-level regression test (TestSimulateV1_LogBlockNumber_MatchesSimulatedHeader) that runs two btctoken.transfer calls in one simulated block and asserts every log's blockNumber and blockHash match the envelope.

tomaszslabon:
I tested manually various simulation scenarios and did not find any more errors.

tomaszslabon (approved):

PR #673, opened 2026-05-05
Add post-Cancun fork-time fields to ChainConfig

lukasz-zimnoch:
References: [MEZO-4013](https://linear.app/thesis-co/issue/MEZO-4013/eip-7702-setcode-transaction-support)
Closes: [MEZO-4007](https://linear.app/thesis-co/issue/MEZO-4007/chain-configuration-extension)

### Introduction

Mezo's EVM module needs to expose every post-Cancun fork-time field that geth `v1.16.9-mezo0` declares on `params.ChainConfig`, so the chain can stage Prague (which carries EIP-7702 SetCode) at genesis and so future activations of Osaka, Amsterdam, Verkle, and the BPO blob-parameter forks can be performed by the EVM module's params alone, without further proto regeneration.

#### `proto/ethermint/evm/v1/evm.proto`

Adds nine new timestamp-fork fields to the `ChainConfig` message, mirroring the existing `shanghai_time` (22) and `cancun_time` (23) shape: `prague_time` (24), `osaka_time` (25), `bpo1_time`..`bpo5_time` (26–30), `amsterdam_time` (31), `verkle_time` (32). The five BPO fields carry an explicit `gogoproto.customname` annotation so the generated Go names match geth (`BPO1Time` etc.) rather than the gogoproto default `Bpo1Time`.

#### `x/evm/types/chain_config.go`

Wires all nine new fields through the existing helpers — `EthereumConfig()` calls `getTimeValue` for each, and `Validate()` calls `validateBlock` for each (geth's `CheckConfigForkOrder` then handles fork ordering).

`DefaultChainConfig()` defaults `PragueTime` to zero, so a fresh genesis activates Prague (and EIP-7702) immediately. The other eight fork times are deliberately left `nil`: the keeper, ante handler, and RPC surface have no support for these forks yet, and defaulting them to zero would silently turn on geth-side jump-table changes (for example the Osaka MODEXP gas schedule). A block comment explains this and notes that activating Osaka or any later fork additionally requires extending `params.DefaultBlobSchedule` (currently only Cancun / Prague / Osaka). Activation on live mainnet and testnet networks will be performed through a planned upgrade handler, not at genesis.

#### `x/evm/types/chain_config_test.go`

Adds `PragueTime: newIntPtr(0)` to the existing "valid" case and nine new "invalid XTime" cases asserting that `Validate()` rejects negative timestamps, mirroring the existing `ShanghaiTime`/`CancunTime` pattern.

#### `x/evm/types/evm.pb.go`

Regenerated via `make proto-gen` (Docker proto-builder 0.11.5).

#### `app/ante/{evm,cosmos,utils}/setup_test.go` and `x/evm/keeper/setup_test.go`

The four `SetupTest` paths that build a chain config without London push every later fork time to `MaxInt64`. With `PragueTime` now defaulting to zero in `DefaultChainConfig`, those configs landed with `CancunTime` at `MaxInt64` and `PragueTime` at `0`, which fails geth's `CheckConfigForkOrder` and panics `InitGenesis`. `PragueTime` is now pushed to `MaxInt64` alongside the existing time fields so the disabled-London path stays consistent with the new default.

### Testing

- `go test -race ./x/evm/types/...` passes, including the nine new validation cases.
- `go test -race ./app/ante/... ./x/evm/keeper/...` passes; the disabled-London setup-test path no longer panics.
- `make build` succeeds; the new fields propagate through generated code without further wiring required.
- `golangci-lint run ./x/evm/types/...` reports zero findings on the touched files. The repo-wide `make lint` has pre-existing typecheck failures in unrelated packages (`precompile/`, `bridge-worker/`, `server/`, `cmd/`, etc.) that are not introduced by this PR.


x/evm/types/chain_config.go
```diff
@@ -173,6 +199,33 @@ func (cc ChainConfig) Validate() error {
...
+	if err := validateBlock(cc.BPO4Time); err != nil {
+		return errorsmod.Wrap(err, "BPO4Time")
+	}
+	if err := validateBlock(cc.BPO5Time); err != nil {
+		return errorsmod.Wrap(err, "BPO5Time")
+	}
+	if err := validateBlock(cc.AmsterdamTime); err != nil {
+		return errorsmod.Wrap(err, "AmsterdamTime")
+	}
+	if err := validateBlock(cc.VerkleTime); err != nil {
```
tomaszslabon: `Verkle/EIP-4762` is an Ethereum-specific feaure and we probably should not (and cannot) support it in Mezo (note that we are leaving `StateDb.AccessEvents()` unimplemented).
I think we should issue an error here if somebody provided a non-nil `VerkleTime`.

lukasz-zimnoch: This must be always determined case by case. In this PR we are just enabling Prague. We add the other forks to the config to ensure parity with geth 1.16 config but that's it.

tomaszslabon (approved):

PR #674, opened 2026-05-05
Add proto messages for SetCode transactions

lukasz-zimnoch:

### Introduction

Adds the on-the-wire shape for EIP-7702 (Set Code Transactions) to mezod's EVM module so the rest of the stack can carry type-`0x04` transactions without translating to/from raw RLP at every layer. No type-`0x04` transaction is yet accepted by the chain — that is gated behind a Prague-active check and the keeper's authorization-processing path is not yet implemented.

#### `SetCodeTx` and `SetCodeAuthorization` proto messages

New protobuf messages in `proto/ethermint/evm/v1/tx.proto` carry the type-`0x04` envelope and per-tuple `(chain_id, address, nonce, v, r, s)` authorization payload, mirroring the shape of `DynamicFeeTx`. Both are registered as `TxData` via `codec.go` and `tx_data.go`, and the type-URL switch in `msg.go::MsgEthereumTxGetSigners` recognizes the new message so signer resolution succeeds for type-`0x04` transactions.

#### Cosmos-side validation

`SetCodeTx.Validate()` rejects malformed type-`0x04` transactions before they reach the keeper:

- A non-empty `to` (no contract creation per EIP-7702) and a non-empty `auth_list`
- Non-nil `ChainID`, `GasTipCap`, `GasFeeCap`; `Amount` may be nil (treated as zero, matching `DynamicFeeTx` and EIP-7702 value semantics); standard gas-cap monotonicity and Mezo chain id allowlist
- Per-tuple wire shape: `auth.ChainID` non-nil and within int256 bounds, `auth.Address` parses, `auth.V`/`R`/`S` within int256 bounds. Canonical-form checks (V in {0,1}, R/S non-zero, length <= 32) and the EIP-7702 0-or-current chain-id rule are deferred to the keeper's authorization processing, where invalid tuples are silently skipped per EIP-7702 — front-loading them at stateless validate would convert per-tuple skips into tx-fatal errors
- Outer `V`/`R`/`S` within int256 bounds so `AsEthereumData`'s `uint256.MustFromBig` conversion stays panic-free on a stored tx

New error sentinels `ErrSetCodeMissingTo` and `ErrSetCodeEmptyAuthList` cover the two strict-rejection cases the EIP forbids.

#### Pre-Prague ante gate

`EthValidateBasicDecorator` rejects type-`0x04` transactions with `ErrTxTypeNotSupported` when `IsPrague` is false at the block time, mirroring the existing pre-London gate for `DynamicFeeTxType`. This holds the wire integration off until the upgrade handler activates Prague on a given network.

#### `newRPCTransaction` switch arm

`simulate_v1.go::newRPCTransaction` panics on `SetCodeTxType` with a TODO pointing at the MEZO-4336 plan; full set-code RPC support lands there.

#### `chain_config_test.go` cleanup

`TestChainConfigValidate` expresses each negative case as a one-field mutation on a shared `validBaselineConfig()` helper instead of restating every `ChainConfig` field per case (~330 lines of duplication removed). New `TestChainConfigEthereumConfig` populates each fork-time field with a distinct sentinel and asserts the corresponding field on the returned `*params.ChainConfig`, so a swapped assignment would surface immediately.

### Testing

`go test ./x/evm/types/... ./app/ante/...` — pass. Specifically:

- `TestNewSetCodeTx`, `TestSetCodeTxValidate` (table cases for every nil/out-of-bound branch including outer V/R/S and per-tuple V/R/S int256 bounds, malformed addresses, auth ChainID nil/zero/non-Mezo-non-zero/out-of-bound, and explicit acceptance cases for canonical-form variants the keeper handles — V[0] > 1, zero R, zero S, non-Mezo non-zero auth chain id), `TestSetCodeTxCopy` (nil-vs-empty AuthList preserved), `TestSetCodeTxAsEthereumData` (covers nil-amount round-trip), `TestSetCodeTxFee`, `TestSetCodeTxAnyRoundTrip` (real wire round-trip via `proto.Marshal`/`proto.Unmarshal` of the `Any` envelope, to ensure `RegisterInterfaces` resolves the type URL)
- `TestPackTxData` extended with `&SetCodeTx{}`
- `TestNewTxDataFromTx_SupportedTypes` extended with the new type
- `TestChainConfigValidate` (refactored) and `TestChainConfigEthereumConfig` (new)
- `TestAnteHandlerSetCodeTxRequiresPrague` pins the pre-Prague ante rejection end-to-end through the full ante chain

`go vet ./x/evm/types/... ./app/ante/evm/...` — clean.

---

### Author's checklist

### Reviewer's checklist

### DIFF
--- a/proto/ethermint/evm/v1/tx.proto
+++ b/proto/ethermint/evm/v1/tx.proto
@@ -150,6 +150,83 @@ message DynamicFeeTx {
   bytes s = 12;
 }
 
+// SetCodeTx is the data of EIP-7702 set-code transactions (type 0x04).
+// `to` must be set (no contract creation) and `auth_list` must be non-empty;
+// proto3 has no field-level required semantics, so these invariants are
+// enforced by Validate() — not by the wire format.
+message SetCodeTx {
+  option (gogoproto.goproto_getters) = false;
+  option (cosmos_proto.implements_interface) = "TxData";
+
+  // chain_id of the destination EVM chain
+  string chain_id = 1 [
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.customname) = "ChainID",
+    (gogoproto.jsontag) = "chainID"
+  ];
+  // nonce corresponds to the account nonce (transaction sequence).
+  uint64 nonce = 2;
+  // gas_tip_cap defines the max value for the gas tip
+  string gas_tip_cap = 3 [ (gogoproto.customtype) = "cosmossdk.io/math.Int" ];
+  // gas_fee_cap defines the max value for the gas fee
+  string gas_fee_cap = 4 [ (gogoproto.customtype) = "cosmossdk.io/math.Int" ];
+  // gas defines the gas limit defined for the transaction.
+  uint64 gas = 5 [ (gogoproto.customname) = "GasLimit" ];
+  // to is the hex formatted address of the recipient
+  string to = 6;
+  // value defines the the transaction amount.
+  string value = 7 [
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.customname) = "Amount"
+  ];
+  // data is the data payload bytes of the transaction.
+  bytes data = 8;
+  // accesses is an array of access tuples
+  repeated AccessTuple accesses = 9 [
+    (gogoproto.castrepeated) = "AccessList",
+    (gogoproto.jsontag) = "accessList",
+    (gogoproto.nullable) = false
+  ];
+  // auth_list carries the EIP-7702 authorization tuples.
+  repeated SetCodeAuthorization auth_list = 10 [
+    (gogoproto.castrepeated) = "AuthorizationList",
+    (gogoproto.jsontag) = "authorizationList",
+    (gogoproto.nullable) = false
+  ];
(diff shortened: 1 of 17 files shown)


proto/ethermint/evm/v1/tx.proto
```diff
@@ -150,6 +150,83 @@ message DynamicFeeTx {
   bytes s = 12;
 }
 
+// SetCodeTx is the data of EIP-7702 set-code transactions (type 0x04).
+// `to` must be set (no contract creation) and `auth_list` must be non-empty;
+// proto3 has no field-level required semantics, so these invariants are
+// enforced by Validate() — not by the wire format.
+message SetCodeTx {
```
tomaszslabon, line 157: @lukasz-zimnoch 
The new transaction type is already registered, so we should also remember to add `SetCodeTx` handling in [msg.go](https://github.com/mezo-org/mezod/blob/set-code-2/x/evm/types/msg.go#L244-L252) in this or a follow-up PR. We could add a TODO here.

lukasz-zimnoch: Added in a5c839d0.

tomaszslabon (approved):

PR #680, opened 2026-05-07
Apply EIP-7702 authorizations in EVM state transition

lukasz-zimnoch:
Depends on: https://github.com/mezo-org/mezod/pull/674

### Introduction

Wires EIP-7702 authorization processing into the EVM keeper's state transition. After this PR, a type-`0x04` transaction that reaches the keeper validates and applies its authorizations (nonce bump, delegation install or clear, target warming, refund accounting), is charged the correct intrinsic gas, and executes its top-level call against the post-authorization state. Authorizations whose target is any precompile — stock Ethereum (0x01..0x12) or a Mezo custom precompile (the `0x7b7c…` range) — are rejected at validate time and silently skipped per EIP-7702's per-tuple rule, since Mezo registers stored facade bytecode at every custom precompile address and a delegation pointing there would actually execute in the authority's context. Self-sponsored authorizations are honored across every consensus and keeper-internal entry point — `EthereumTx`, `EthCall`, `EstimateGas`, `TraceTx`, `traceTx`, `TraceBlock`, and `SimulateV1`. Type-`0x04` is still gated by the pre-Prague ante check from #674 and the user-visible activation upgrade is out of scope here, so no end-user surface accepts SetCode txs yet — but the consensus path is now correct end-to-end inside the keeper.

#### Authorization processing helpers

A new `x/evm/keeper/set_code_authorization.go` reimplements geth's unexported `validateAuthorization` and `applyAuthorization` on top of mezod's `*statedb.StateDB`. Rebuilding a geth `stateTransition` to reach the upstream methods would mean abandoning mezod's `MinGasMultiplier`, custom precompile registry, and simulate-driver paths, so the helpers run inline against the same StateDB the rest of the tx commits through. Pin references on each function are anchored by function name + commit SHA against `mezo-org/go-ethereum@v1.16.9-mezo0` so future geth bumps re-audit them without the brittleness of upstream line numbers.

`validateSetCodeAuthorization` enforces the EIP-7702 invariants (chain id is `0` or current; nonce + 1 does not overflow; signature recovers; authority's existing code is empty or already a delegation; authority's nonce matches `auth.Nonce`) and adds the recovered authority to the access list before the destination-code and nonce checks, mirroring upstream's deliberate ordering so a failed tuple still warms its authority. It also enforces a Mezo-specific rule: any authorization whose target sits in `evm.Precompiles()` (the union of fork-active stock precompiles and Mezo customs) is rejected with a dedicated sentinel error before signature recovery — so a precompile-target tuple does not pay ECDSA-recovery cost and is not added to the access list. `applySetCodeAuthorization` calls the validator, bumps the authority's nonce, installs `0xef0100 || target` (or clears the code when `auth.Address == 0x0`), and refunds `params.CallNewAccountGas - params.TxAuthTupleGas` to the StateDB if the authority already exists in state. Per-tuple validation errors are sentinel `error` values; the auth-loop caller emits them at debug level with the auth index and the rejection reason and otherwise discards them per EIP-7702's "invalid tuples are skipped" rule.

The per-tuple loop and the post-loop access-list warming of `msg.To`'s resolved delegation target are wrapped in a single `applySetCodeAuthorizations` helper. The `isPrecompile` predicate is threaded through the loop and the per-tuple validator/applier as a `func(common.Address) bool`; the call site in `applyMessageWithConfig` snapshots `evm.Precompiles()` once before invoking the loop so the upstream `maps.Clone` inside `Precompiles()` runs once per message rather than once per tuple.

#### State-transition wiring

`applyMessageWithConfig` runs `applySetCodeAuthorizations` on the call path, between intrinsic-gas charging and `evm.Call`, against the same `*statedb.StateDB` the call uses. After the per-tuple loop, if `msg.To`'s code parses as a delegation marker, the resolved target is added to the access list so a delegation just installed by this same tx becomes warm before the call lands — replicating geth's post-loop warming. A defense-in-depth guard at function entry mirrors geth's preCheck pair: when `SetCodeAuthorizations != nil`, `To == nil` is rejected with `core.ErrSetCodeTxCreate` and a length-0 slice is rejected with `core.ErrEmptyAuthList`, so simulate / RPC paths that build a `core.Message` directly cannot bypass the EIP-7702 envelope rules. The bridge-internal `ExecuteContractCall` leaves `SetCodeAuthorizations` nil — a non-nil empty slice would now trip the keeper-level guard — with an inline note pinning the reason.

The sender-nonce bump that geth performs inline before the auth loop is supplied here by `EthIncrementSenderSequenceDecorator` on the consensus path. Keeper-internal entry points bypass the ante and mirror the bump at the entry point so self-sponsored authorizations validate against the same post-bump nonce as the consensus path. `EthCall`, `TraceTx`'s predecessor loop, and `traceTx`'s singular call site (which serves both `TraceTx`'s target dispatch and `TraceBlock`'s per-tx loop) call a shared `bumpAccNonce` helper in `x/evm/keeper/statedb.go` that advances the AccountKeeper sequence; `EstimateGasInternal` calls the same helper inside its per-iteration `CacheContext`. `SimulateV1`'s per-call inner driver still uses `sdb.SetNonce` directly because its StateDB caches state objects across calls, and an AccountKeeper write would not be visible through the cached object. `SimulateV1`'s post-call `sdb.SetNonce` is removed; the per-call pre-bump subsumes that advance and avoids clobbering the auth-loop's own bump on self-sponsored auths.

Because `bumpAccNonce` writes the new sequence directly to the shared `sdk.Context` before `ApplyMessageWithConfig` runs, a state-transition error returned from apply would leave the sender-nonce advance behind without the matching EVM state. `TraceTx`'s predecessor loop and `TraceBlock`'s per-tx loop now abort the entire RPC request on such an error rather than swallowing it via `continue` or recording it in `TxTraceResult.Error` and proceeding — both prior shapes would let the leaked bump corrupt every later iteration's baseline. The abort matches go-ethereum's sequential `debug_traceBlock` (`eth/tracers/api.go`), which returns on the first state-transition failure rather than producing partial per-tx results. EVM-level reverts continue to surface as `res.Failed() == true` with `err == nil` and never reach the abort branch.

#### JSON-RPC ingress

`x/evm/types/TransactionArgs` gains an `authorizationList` field carrying `[]ethtypes.SetCodeAuthorization` with the geth-canonical JSON tag. `ToMessage` populates `core.Message.SetCodeAuthorizations` from it, and `EstimateGasInternal`'s per-iteration message rebuild re-sets the slice so the gas binary-search runs the auth loop. Wallets calling `eth_call`, `eth_estimateGas`, or `debug_traceCall` with a SetCodeTx payload now see the per-tuple `CallNewAccountGas` surcharge in the gas estimate and observe the auth side-effects in the simulate result.

#### Intrinsic-gas plumbing

`GetEthIntrinsicGas` and `VerifyFee` thread the authorization list into `core.IntrinsicGas`, so each tuple adds `params.CallNewAccountGas` (25000) to the intrinsic charge. `VerifyFee` reads the list through a new `GetAuthorizationList()` accessor on the `TxData` interface; `LegacyTx`, `AccessListTx`, and `DynamicFeeTx` return `nil` and behave identically to before. The `TODO (geth-upgrade)` markers in `gas.go` and `fees.go` are removed.

#### Tests

Differential keeper tests cover the new helpers and the wired-in auth loop: validation sentinels with geth-aligned access-list ordering, success paths (install / clear / rotate / target-is-contract / authority-was-delegated / nilChainID / refund accounting), end-to-end `ApplyMessage` drives (install, duplicate-authority-last-wins, post-loop target warming, invalid-tuple silent skip, auth survives top-level revert, rejects auth list with nil To, rejects empty auth list, self-sponsored no-double-bump), and entry-point coverage for `EthCall`, `EstimateGas`, `TraceBlock`, `TraceTx` predecessor, and `SimulateV1`. The precompile-target gate is pinned by `TestApplySetCodeAuthorization_RejectsPrecompile` (table-driven over stock 0x01 / 0x09 / 0x0a, custom btctoken, custom assetsbridge, plus a zero-address negative control) and `TestApplySetCodeAuthorizations_PrecompileTupleSilentlySkipped` (loop level: bad tuple skipped, good tuple applies, post-loop msg.To warming still runs). Gas-plumbing pins (`TestGetEthIntrinsicGasSetCodeTx`, `TestVerifyFeeSetCodeAuthList`, `TestSetCodeTx_AuthListRoundTripCount`) lock down intrinsic charge + Cosmos↔core auth-count round-trip. `BenchmarkApplyMessageWithSetCodeTx` parametrises on `N ∈ {0, 1, 16, 64}` with each tuple signed by a fresh ECDSA key so per-tuple ECDSA recovery cost is not artificially shared via geth's signer cache.

#### Specification

`.claude/MEZO-4013-eip7702-set-code/spec.md` documents the consensus-vs-keeper-internal sender-nonce bump asymmetry and adds a `Post-loop call-target warming` bullet describing the `msg.To` delegation-target warming. The precompile-target rejection is recorded as Mezo-specific divergence #4, and the previous "Custom precompiles unaffected" key decision is replaced by one that records the rejection rule and points to the divergence section. `roadmap.md` references this PR.

### Testing

`go test -race -count=1 ./x/evm/keeper/... ./x/evm/types/...` — pass.

`make lint` — clean.

`./tests/system/system-tests.sh` against a fresh localnode — 421 passing, 0 failing.

---

### Author's checklist

### Reviewer's checklist


x/evm/keeper/state_transition.go
```diff
@@ -506,6 +506,14 @@ func (k *Keeper) applyMessageWithConfig(
 		gasUsed uint64
 	)
 
+	// EIP-7702 forbids combining a non-empty auth list with contract creation.
+	// SetCodeTx.Validate and the ante reject this on the consensus path, but
+	// simulate / RPC ingress can construct a core.Message that bypasses both
+	// — enforce the invariant directly so the create branch is unreachable.
+	if len(msg.SetCodeAuthorizations) > 0 && msg.To == nil {
```
tomaszslabon: Maybe this should be:
```
if msg.SetCodeAuthorizations != nil {
	if msg.To == nil {
		return nil, nil, errorsmod.Wrap(core.ErrSetCodeTxCreate, "apply message")
	}
	if len(msg.SetCodeAuthorizations) == 0 {
		return nil, nil, errorsmod.Wrap(core.ErrEmptyAuthList, "apply message")
	}
}
```

In [go-ethereum](https://github.com/mezo-org/go-ethereum/blob/v1.16.9-mezo0/core/state_transition.go#L401-L407), a non-nil but empty authorization list is rejected with `core.ErrEmptyAuthList`.

x/evm/types/tx_args.go
```diff
@@ -244,16 +247,17 @@ func (args *TransactionArgs) ToMessage(globalGasCap uint64, baseFee *big.Int) (c
...
+		To:                    args.To,
+		Nonce:                 nonce,
+		Value:                 value,
+		GasLimit:              gas,
+		GasPrice:              gasPrice,
+		GasFeeCap:             gasFeeCap,
+		GasTipCap:             gasTipCap,
+		Data:                  data,
+		AccessList:            accessList,
+		SetCodeAuthorizations: args.AuthorizationList,
```
tomaszslabon, line 260: We probably should update the `TransactionArgs.ToTransaction()` function to handle cases when `AuthorizationList` is not nil. In that cases we should `SetCodeTx`. We could do something like this:
```
        var data TxData
        switch {
+       case args.AuthorizationList != nil:
+               al := AccessList{}
+               if args.AccessList != nil {
+                       al = NewAccessList(args.AccessList)
+               }
+
+               data = &SetCodeTx{
+                       To:        to,
+                       ChainID:   &chainID,
+                       Nonce:     nonce,
+                       GasLimit:  gas,
+                       GasFeeCap: &maxFeePerGas,
+                       GasTipCap: &maxPriorityFeePerGas,
+                       Amount:    &value,
+                       Data:      args.GetData(),
+                       Accesses:  al,
+                       AuthList:  NewAuthorizationList(args.AuthorizationList),
+               }
        case args.MaxFeePerGas != nil:
                al := AccessList{}
```

This would be consistent with how [geth selects](https://github.com/mezo-org/go-ethereum/blob/v1.16.9-mezo0/internal/ethapi/transaction_args.go#L503-L538) `SetCodeTx` for non-nil `AuthorizationList`.

x/evm/keeper/state_transition.go
lukasz-zimnoch: Adopted in a4ff1004. ExecuteContractCall had to flip its SetCodeAuthorizations from an empty slice to nil at the same time — the stricter guard would otherwise have errored every internal call.

tomaszslabon:
A comment related to EIP-7702: we recently added [AuthorizationList](https://github.com/mezo-org/mezod/blob/set-code-3/x/evm/types/tx_args.go#L56-L57) field to the `TransactionArgs` struct.

I think we should pass it into the temporary `callArgs` in `SetTxDefaults()`. Otherwise, gas estimation for args with `authorizationList` but no explicit `gas` ignores the EIP-7702 auth-list intrinsic gas and may set the gas limit too low.

Suggested change:
```
@@ -285,6 +285,7 @@ func (b *Backend) SetTxDefaults(args evmtypes.TransactionArgs) (evmtypes.Transac
                        Value:                args.Value,
                        Data:                 input,
                        AccessList:           args.AccessList,
+                       AuthorizationList:    args.AuthorizationList,
                        ChainID:              args.ChainID,
                        Nonce:                args.Nonce,
                }
```

x/evm/types/tx_args.go
lukasz-zimnoch: This will be added as part of phase 5 (see `roadmap.md`).

EDIT: Added in https://github.com/mezo-org/mezod/pull/683

lukasz-zimnoch:
> A comment related to EIP-7702: we recently added [AuthorizationList](https://github.com/mezo-org/mezod/blob/set-code-3/x/evm/types/tx_args.go#L56-L57) field to the `TransactionArgs` struct.
> 
> I think we should pass it into the temporary `callArgs` in `SetTxDefaults()`. Otherwise, gas estimation for args with `authorizationList` but no explicit `gas` ignores the EIP-7702 auth-list intrinsic gas and may set the gas limit too low.
> 
> Suggested change:
> 
> ```
> @@ -285,6 +285,7 @@ func (b *Backend) SetTxDefaults(args evmtypes.TransactionArgs) (evmtypes.Transac
>                         Value:                args.Value,
>                         Data:                 input,
>                         AccessList:           args.AccessList,
> +                       AuthorizationList:    args.AuthorizationList,
>                         ChainID:              args.ChainID,
>                         Nonce:                args.Nonce,
>                 }
> ```

This will be added as part of phase 5 (see `roadmap.md`).

EDIT: Added in https://github.com/mezo-org/mezod/pull/683

x/evm/keeper/grpc_query.go
```diff
@@ -516,6 +510,11 @@ func (k Keeper) TraceTx(c context.Context, req *types.QueryTraceTxRequest) (*typ
 		if err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
+
+		if err := k.bumpAccNonce(ctx, msg.From); err != nil {
```
tomaszslabon, line 514: `bumpAccNonce(ctx, msg.From)` mutates `ctx` before `ApplyMessageWithConfig()`. If apply returns an error, the nonce bump remains in the trace context. In `TraceBlock`, that can affect replay of following txs.

Maybe we could use `CacheContext` (in two places), e.g.:
```
tslabon@MacBook-Pro-Tomasz-2 ~/projekt/mezod % git diff
diff --git a/x/evm/keeper/grpc_query.go b/x/evm/keeper/grpc_query.go
index a1bf948a..440e42e2 100644
--- a/x/evm/keeper/grpc_query.go
+++ b/x/evm/keeper/grpc_query.go
@@ -511,14 +511,16 @@ func (k Keeper) TraceTx(c context.Context, req *types.QueryTraceTxRequest) (*typ
                        return nil, status.Error(codes.Internal, err.Error())
                }
 
-               if err := k.bumpAccNonce(ctx, msg.From); err != nil {
+               tmpCtx, write := ctx.CacheContext()
+               if err := k.bumpAccNonce(tmpCtx, msg.From); err != nil {
                        return nil, status.Error(codes.Internal, err.Error())
                }
 
-               rsp, _, err := k.ApplyMessageWithConfig(ctx, WrapMessageWithSource(*msg, ethTx), tracer, true, cfg, txConfig)
+               rsp, _, err := k.ApplyMessageWithConfig(tmpCtx, WrapMessageWithSource(*msg, ethTx), tracer, true, cfg, txConfig)
                if err != nil {
                        continue
                }
+               write()
                txConfig.LogIndex += uint(len(rsp.Logs))
        }
 
@@ -700,14 +702,18 @@ func (k *Keeper) traceTx(
                }
        }()
 
-       if err := k.bumpAccNonce(ctx, msg.From); err != nil {
+       execCtx, write := ctx.CacheContext()
+       if err := k.bumpAccNonce(execCtx, msg.From); err != nil {
                return nil, 0, status.Error(codes.Internal, err.Error())
        }
 
-       res, _, err := k.ApplyMessageWithConfig(ctx, WrapMessageWithSource(*msg, tx), tracer, commitMessage, cfg,
 txConfig)
+       res, _, err := k.ApplyMessageWithConfig(execCtx, WrapMessageWithSource(*msg, tx), tracer, commitMessage, 
cfg, txConfig)
        if err != nil {
                return nil, 0, status.Error(codes.Internal, err.Error())
        }
+       if commitMessage {
+               write()
+       }
 
        var result interface{}
        result, err = tracer.GetResult()
```

We already use this pattern in normal tx execution to avoid leaking partial state when later steps fail:
https://github.com/mezo-org/mezod/blob/set-code-3/x/evm/keeper/state_transition.go#L291-L303

lukasz-zimnoch: Went stronger than the suggestion: bubbling the apply error up to abort the entire trace request, matching go-ethereum's sequential `debug_traceBlock` (`eth/tracers/api.go`) which returns on the first state-transition failure rather than emitting partial per-tx results. `TraceBlock`'s per-tx loop gets the same treatment. With apply errors no longer swallowed, the leaked bump has no later iteration to corrupt, so `CacheContext` becomes unnecessary at this site. EVM-level reverts surface as `res.Failed() == true` with `err == nil` and never enter the abort branch.

Commit: 9c28c1363d473e53a4dbf3be6cd4554c95065f40

tomaszslabon (approved):

PR #683, opened 2026-05-11
Wire EIP-7702 SetCode through the Ethereum JSON-RPC surface

lukasz-zimnoch:
Depends on: https://github.com/mezo-org/mezod/pull/681

### Introduction

Wires EIP-7702 type-`0x04` transactions through every public Ethereum JSON-RPC entry point. After this PR, an integrator can submit a SetCodeTx through `eth_sendRawTransaction` or `eth_sendTransaction`, observe it through `eth_getTransactionByHash`, `eth_getBlockByNumber`, `eth_getTransactionReceipt`, and pass an `authorizationList` to `eth_call` / `eth_estimateGas` for accurate intrinsic-gas estimation. Receipts surface `effectiveGasPrice` on SetCodeTx the same way geth does for any post-1559 tx type. Authorization processing in the keeper now rejects pre-Prague messages carrying an auth list with `ErrTxTypeNotSupported`, matching the SetCodeTxType rejection in the consensus-path ante and closing the misleading acceptance / rejection split for simulate paths that bypass the ante.

#### `x/evm/types/tx_args.go`

`TransactionArgs` gains an `AuthorizationList []ethtypes.SetCodeAuthorization` field (JSON-tagged `authorizationList`, without `omitempty`, so an explicitly empty list survives the `json.Marshal` step in `Backend.DoCall` / `SetTxDefaults` and reaches downstream validation as a non-nil slice instead of being silently elided). `ToTransaction` mirrors upstream geth's `internal/ethapi/transaction_args.go` shape: a prologue switch picks a `usedType` from the populated fields (`AuthorizationList` → `SetCodeTxType`, `MaxFeePerGas` → `DynamicFeeTxType`, `AccessList` → `AccessListTxType`, else `LegacyTxType`), then a separate override forces `LegacyTxType` when `args.GasPrice` is set, and dispatch runs on `usedType`. The `nil`-vs-`len` predicate on the auth-list branch mirrors geth: an empty non-nil list still routes to type-`0x04` so downstream validation can reject it per EIP-7702. The `GasPrice` override applies uniformly — a request that mixes `GasPrice` with an auth list, `MaxFeePerGas`, or `AccessList` lands in `LegacyTx`, matching upstream's "use legacy if gasprice is provided" contract. `ToMessage` forwards the auth list into `core.Message.SetCodeAuthorizations` so the keeper's authorization-application path sees it on simulate/call entrypoints. `String()` now summarises each auth entry as `(chainId=X address=Y nonce=Z)` rather than printing only the length.

#### `rpc/types/utils.go` + `rpc/types/types.go`

`NewRPCTransaction` gains a `SetCodeTxType` arm that emits `authorizationList`, the access list, the chain ID, and the 1559 fee fields via a shared `setEffectiveGasPrice` helper that also serves the existing `DynamicFeeTxType` arm. `RPCTransaction` gains the `AuthorizationList []ethtypes.SetCodeAuthorization` field. The slice returned by geth's `tx.SetCodeAuthorizations()` aliases the underlying transaction's auth list and is documented as read-only at the assignment site.

#### `rpc/backend/tx_info.go`

`GetTransactionReceipt` now writes `effectiveGasPrice` for any `txData` whose `TxType() >= ethtypes.DynamicFeeTxType`, replacing the previous DynamicFeeTx-only branch. SetCodeTx, AccessListTx, and any future post-1559 type ride the same path through the existing `TxData.EffectiveGasPrice` interface method.

#### `rpc/backend/call_tx.go`

The signer-resolution path was already type-agnostic via `MakeSigner`; only the receipt and tx-args paths needed wiring. A single-line tightening keeps the SetCodeTx case from being silently dropped through the cosmos tx encoder boundary.

#### `x/evm/keeper/state_transition.go`

`applyMessageWithConfig` now treats a non-Prague auth list as a top-level fatal alongside the existing auth-with-create and empty-auth-list rejections in the function prologue. The check returns `ErrTxTypeNotSupported`, matching the `SetCodeTxType` rejection in `setup_ctx.go`'s ante so simulate / `eth_call` / `eth_estimateGas` (which bypass the ante) surface the same error rather than silently dropping the auth list. With the rejection in place the call branch invokes `applySetCodeAuthorizations` unconditionally — pre-Prague the auth list is guaranteed nil and no delegation marker can yet exist at `msg.To`, so the helper is a no-op. The `rules` computation that fed the old call-site Prague gate is performed once at the top of the function and reused for the `IsBerlin` access-list `Prepare`.

#### Tests

- `x/evm/types/tx_args_test.go` — field-by-field assertions for the SetCodeTx branch (both populated and `nil` access-list variants), the `gasPrice + authorizationList` short-circuit, the empty-non-nil-auth-list case, plus updated `String()` expectations. `TestAuthorizationListJSONRoundTrip` pins that an explicit `authorizationList: []` survives a `json.Marshal` / `json.Unmarshal` round-trip and that an absent field still unmarshals to a nil slice.
- `rpc/types/utils_test.go` — `TestNewRPCTransaction_SetCodeTx` and `TestNewRPCTransaction_DynamicFeeTx` cover JSON marshalling of `authorizationList`, the `effectiveGasPrice` formula with and without a base fee, and the `tip+baseFee > feeCap` cap.
- `rpc/backend/tx_info_test.go` — `TestGetTransactionReceiptSetCodeTxEffectiveGasPrice` indexes a real SetCodeTx, fetches the receipt, and pins that `effectiveGasPrice`, `type`, and `from` are all correct.
- `rpc/backend/call_tx_test.go` — `TestSendRawTransactionSetCodeTx` asserts signer + authority recovery on the raw-send path; `TestSetTxDefaultsPropagatesAuthorizationList` + `TestSetTxDefaultsLeavesNilAuthorizationList` pin nil-vs-populated `AuthorizationList` propagation through `SetTxDefaults` into `EstimateGas`.
- `tests/system/test/Eip7702SendRawTx.test.ts` — a hardhat suite that funds a fresh EOA, signs a self-sponsored EIP-7702 authorization, submits the type-`0x04` envelope through `eth_sendRawTransaction`, and asserts the receipt status, RPC tx type/authorizationList shape, and the resulting `0xef0100||target` delegation designator on the EOA's code field.

### Testing

`go build ./...` clean. `go test -race -timeout 5m ./rpc/backend/ ./rpc/types/ ./x/evm/types/ ./x/evm/keeper/` passes. The new hardhat suite (`./tests/system/system-tests.sh Eip7702SendRawTx`) passes against a localnode (`make localnode-bin-start`, chain id `0x7b7b`/`31611`).

---

### Author's checklist

### Reviewer's checklist

### DIFF
--- a/.claude/MEZO-4013-eip7702-set-code/roadmap.md
+++ b/.claude/MEZO-4013-eip7702-set-code/roadmap.md
@@ -26,6 +26,17 @@
   `EthAccountVerificationDecorator` while genuine contract code is
   still rejected.
   PR: [#681](https://github.com/mezo-org/mezod/pull/681).
+- **Phase 5.** JSON-RPC surface: `TransactionArgs` gains
+  `authorizationList`, the formatting layer (`NewRPCTransaction`,
+  `RPCTransaction`) emits the field and the type-`0x04` envelope on
+  subscriptions and historical lookups, receipts surface
+  `effectiveGasPrice` for SetCodeTx via the existing
+  `TxData.EffectiveGasPrice` interface, and the keeper's
+  `applyMessageWithConfig` gates `applySetCodeAuthorizations` on
+  `rules.IsPrague` so simulate / `eth_call` / `eth_estimateGas` match
+  consensus on a pre-Prague chain. End-to-end coverage via a Hardhat
+  `Eip7702SendRawTx` system test.
+  PR: [#683](https://github.com/mezo-org/mezod/pull/683).
 
 ## Status of prerequisites
 
--- a/rpc/backend/call_tx.go
+++ b/rpc/backend/call_tx.go
@@ -285,6 +285,7 @@ func (b *Backend) SetTxDefaults(args evmtypes.TransactionArgs) (evmtypes.Transac
 			Value:                args.Value,
 			Data:                 input,
 			AccessList:           args.AccessList,
+			AuthorizationList:    args.AuthorizationList,
 			ChainID:              args.ChainID,
 			Nonce:                args.Nonce,
 		}
--- a/rpc/backend/call_tx_test.go
+++ b/rpc/backend/call_tx_test.go
@@ -9,6 +9,9 @@ import (
 	sdkmath "cosmossdk.io/math"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/holiman/uint256"
+
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/mezo-org/mezod/rpc/backend/mocks"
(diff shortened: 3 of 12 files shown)


tomaszslabon:
@lukasz-zimnoch

I tested EIP-7702 SetCode txs locally. The real `eth_sendRawTransaction` path looks OK: a valid type `0x04` tx succeeds, and a type `0x04` tx with an empty auth list is rejected.

I found one edge case in `eth_call` though:
```
curl -sS http://localhost:8545 \
  -H 'Content-Type: application/json' \
  --data '{
    "jsonrpc":"2.0",
    "id":1,
    "method":"eth_call",
    "params":[{
      "from":"0x0000000000000000000000000000000000000001",
      "to":"0x0000000000000000000000000000000000000002",
      "authorizationList":[]
    },"latest"]
  }' | jq
  ```
Current response:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": "0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
}
```
I think this should return the empty-auth-list error instead, because `authorizationList` is present but empty.

The issue seems to be that `AuthorizationList` currently uses `omitempty` in `TransactionArgs`, so `authorizationList: []` can be dropped during internal JSON marshaling:
https://github.com/mezo-org/mezod/blob/set-code-5/x/evm/types/tx_args.go#L56-L57

Geth does not use `omitempty` for this field:
https://github.com/mezo-org/go-ethereum/blob/v1.16.9-mezo0/internal/ethapi/transaction_args.go#L71-L72

Suggested fix: remove `omitempty` from the `AuthorizationList` JSON tag.

lukasz-zimnoch:
Fixed in 1bffed2b. Dropped `omitempty` from the `AuthorizationList` JSON tag so `authorizationList: []` survives the internal `json.Marshal` step in `Backend.DoCall` / `SetTxDefaults` and the keeper's empty-list check fires on `eth_call` and `eth_estimateGas` too. Added a JSON round-trip regression test in `tx_args_test.go`.

tomaszslabon (approved):

PR #684, opened 2026-05-11
Add per-validator bridge-out attestation count metric

Shadowfiend:
## Summary

- Adds `bridge_val_attestation_count` Prometheus counter to the metrics-scraper, tracking per-bridge-validator attestation activity on the MezoBridge contract
- Exposes the `MezoBridgeAssetsUnlockAttested` event type from the portal bindings package

## Motivation

During the 2026-05-09 bridge-out halt, 2 of 9 bridge validators had silently stopped attesting for an extended period. The system was running at exactly 7/9 quorum with zero fault tolerance. When a 3rd validator dropped off (likely due to ETH balance exhaustion), quorum broke and all bridge-outs from sequence 8068 onward stalled for 12+ hours.

The existing `pending_assets_unlocked` metric only fires _after_ quorum is already lost — a lagging indicator. This new metric enables detecting non-participating validators _before_ they cause a halt.

## Intended usage

Example Grafana alert (warning severity):

```promql
increase(bridge_val_attestation_count[6h]) == 0
  AND ON ()
max(increase(bridge_val_attestation_count[6h])) > 0
```

This fires when a validator has zero attestations in 6 hours while others are actively attesting. It won't trigger during quiet periods with no bridge-out activity.

## Test plan

- [ ] Verify `go vet ./metrics-scraper/...` passes (confirmed locally)
- [ ] Deploy to staging metrics-scraper and confirm the counter appears in Prometheus with per-validator labels
- [ ] Verify counter increments when bridge-out attestations occur on testnet
- [ ] Configure Grafana alert rule using the PromQL above

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/metrics-scraper/bridge_monitoring.go
+++ b/metrics-scraper/bridge_monitoring.go
@@ -30,6 +30,9 @@ const (
 	// of the attestation process on Ethereum. Missing those events could result
 	// in false positives in the pending_assets_unlocked metric.
 	ethereumAssetsUnlockedConfirmedLookBackBlocks = uint64(100800) // ~14 days (assuming 1 block per 12s)
+	// Look-back period for AssetsUnlockAttested events used to initialize
+	// per-validator attestation counts on first poll.
+	ethereumAssetsUnlockAttestedLookBackBlocks = uint64(50400) // ~7 days (assuming 1 block per 12s)
 )
 
 var pendingAssetsUnlockedCache = map[string]bool{} // unlock seqno -> bool
@@ -40,6 +43,7 @@ type ethereumChain struct {
 	tbtcBankAddress common.Address
 
 	assetsUnlockedConfirmedCheckpointHeight uint64
+	assetsUnlockAttestedCheckpointHeight    uint64
 }
 
 type mezoChain struct {
@@ -284,6 +288,15 @@ func pollBridgeData(
 	err = outflowLimitSaturation(
 		mezoChainID,
 		mezo,
+	)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	err = bridgeValAttestationCounts(
+		ctx,
+		mezoChainID,
+		ethereum,
 	)
 	if err != nil {
 		errs = append(errs, err)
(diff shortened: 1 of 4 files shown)


metrics-scraper/bridge_monitoring.go
```diff
@@ -698,3 +711,97 @@ func outflowLimitSaturation(
...
+
+	// Get all registered bridge validators so we emit a 0 for validators
+	// that haven't attested at all.
+	validatorsCount, err := ethereum.mezoBridge.BridgeValidatorsCount()
+	if err != nil {
+		return fmt.Errorf("failed to get bridge validators count: [%w]", err)
+	}
+
+	for i := uint64(0); i < validatorsCount.Uint64(); i++ {
+		//nolint:gosec
```
Shadowfiend, line 767: Not completely sure this is right.

metrics-scraper/bridge_monitoring.go
```diff
@@ -698,3 +711,97 @@ func outflowLimitSaturation(
...
+	if err != nil {
+		return fmt.Errorf(
+			"failed to get AssetsUnlockAttested events on Ethereum chain: [%w]",
+			err,
+		)
+	}
+
+	ethereum.assetsUnlockAttestedCheckpointHeight = endHeight + 1
+
+	counts := make(map[common.Address]int)
```
lukasz-zimnoch, line 754: Note that this logic will keep the counters for removed bridge validators and raise false-positive alerts for them. I explored some options to kick removed vals from the counter but they are quite complex comparing to the current solution. Not sure they're worth it given the fact bridge val removal is a rare event and we can suppress specific alerts on Grafana. I think we can merge as is and re-visit if needed.

lukasz-zimnoch:
I've built and pushed a new `mezo/metrics-scraper` Docker image. Bumped our deployments here: ca229c574922eb27a32fe8102df5950ccefc696e

lukasz-zimnoch (approved):

PR #685, opened 2026-05-12
End-to-end test coverage for EIP-7702 SetCode transactions

lukasz-zimnoch:
Depends on: https://github.com/mezo-org/mezod/pull/683

### Introduction

After this PR, mezod's EIP-7702 surface has end-to-end test coverage that exercises every documented spec path, pins every Mezo-specific divergence as a regression tripwire, and protects the `SetCodeTx` proto unmarshaler against arbitrary input. The custom KV indexer's tx-type-agnostic indexing path is also explicitly covered for type-`0x04` transactions, which previously fell through `types.NewTx` (no `SetCodeTx` arm) and reached the indexer only via a single-case smoketest. That smoketest is superseded by the new compliance suite and removed.

#### `tests/system/test/Eip7702_SpecCompliance.test.ts` (new)

Seventeen cases driven by `eth_sendRawTransaction` and read back through the RPC surface, asserting that mezod matches the go-ethereum Prague reference on every EIP-7702 path:

- Delegation install (RPC envelope shape, status, `authorizationList` echoed back, target storage written through the EOA), rotation between two targets while preserving storage, and clearing with `address = 0x0` followed by re-delegation of the same authority — the post-clear re-delegation pin documents that storage is intentionally NOT wiped on clear.
- Self-sponsored authorizations (authority == sender) including the pre-bump-nonce silent-skip case and the no-loop self-delegation case. Both flow through the keeper's `applyMessageWithConfig` post-bump nonce path.
- Single-level resolution: `A -> B -> T` does not collapse to `A -> T`. Inherited verbatim from the geth EVM via the `mezo-org/go-ethereum v1.16.9-mezo0` bump.
- Intrinsic-gas accounting: `CallNewAccountGas - TxAuthTupleGas` (12 500) is refunded when the authority pre-exists in state, and the absolute per-fresh-authority charge (`CallNewAccountGas = 25 000`) is pinned by holding the workload constant and varying only the auth-list length between two type-`0x04` txs.
- Silent-skip per the EIP for duplicated tuples, wrong-chain-id tuples, and tuples with a corrupted signature. The any-chain id (`chainId = 0`) acceptance path is asserted alongside.
- Sender-side EIP-3607 exemption: a delegated EOA submits a normal `DynamicFee` tx and it lands, observing the ante carve-out from PR #681 from the user's seat.
- One-level resolution under `CALL` / `STATICCALL` / `DELEGATECALL` issued by a contract entry point (`Eip7702Caller.sol`).
- `EXTCODECOPY` / `EXTCODESIZE` / `EXTCODEHASH` on a delegated authority return the raw 23-byte `0xef0100 || target` designator, as required by the EIP for on-chain delegation detection.
- EIP-2929 cold-vs-warm CALL gas accounting: a contract issues two sequential CALLs to a delegated EOA in one tx; the first call observes both the authority and the delegate target as cold, the second as warm, and the per-call gas spread (measured via `gasleft()` and emitted as a `CallGas` event) lands in the predicted `[4 800, 5 200]` window — pinning that delegation resolution warms BOTH addresses in the post-loop access-list pass.

Each case builds a fresh sponsor and authority so cases are order-independent. Receipts are polled because `HardhatEthersProvider` does not implement `waitForTransaction`. Gas-related assertions target deltas, not absolute `gasUsed`, because mezod's `MinGasMultiplier` floor and the EIP-3529 refund cap make byte-level gas comparisons brittle.

#### `tests/system/test/Eip7702_MezoDivergence.test.ts` (new)

Four tripwires for the divergences enumerated in `spec.md`. If one of these starts failing, mezod has drifted toward the EIP — investigate before flipping the assertion green.

- Two sequential type-`0x04` transactions from the same authority both land in successive blocks, and two authorization tuples from the same authority inside a single transaction both apply. Geth's txpool reserves the authority across both surfaces; mezod runs on CometBFT without a peer-to-peer txpool of that kind, so neither surface is hardened.
- An authorization whose target is any precompile — every stock precompile (`0x01..0x0a`) and every Mezo custom precompile registered in `x/evm/types/precompile.go` (`0x7b7c…0000` BTCToken through `0x7b7c…0015` PriceOracle) — is rejected at validate time and the rest of the tx proceeds. The clear-delegation path (`target = 0x0`) is unaffected. The custom-precompile list is held inline with a "keep in sync with `x/evm/types/precompile.go`" comment so config drift fails loudly here rather than silently in production.
- Type-`0x04` sender recovery rides on geth's `PragueSigner`; no Mezo-specific signer adapter is introduced. Asserted by recovering the sender from the RPC tx echo and comparing against the sponsor address.

#### `tests/system/test/helpers/eip7702.ts`, `tests/system/contracts/Eip7702*.sol`, `tests/system/deploy/12_deploy_eip7702_fixtures.ts` (new)

Shared infrastructure for the two suites: fresh-wallet funding, authorization signing through `Wallet.authorize`, type-`0x04` envelope submission and receipt polling, delegation-designator construction (`expectedDelegationCode`). Two delegation targets (`Eip7702TargetV1/V2.sol`) with per-version code so rotation cases can observe the switch through `EXTCODE*` and storage. `Eip7702Caller.sol` is the contract entry point for `CALL` / `STATICCALL` / `DELEGATECALL` resolution cases; `Eip7702ExtCodeReader.sol` is the contract entry point for the designator-opacity cases. The `hardhat-deploy` script wires all four under the `eip7702-fixtures` tag.

#### `tests/system/package.json` (bump)

`ethers` devDependency moves from `^6.13.5` to `^6.16.0`. The 6.16 line is where `Wallet.authorize()` lands as a first-class API; the 6.13 line does not expose the EIP-7702 signing surface, so the helper module would otherwise have to hand-roll RLP and secp256k1.

#### `x/evm/types/set_code_tx_fuzz_test.go` (new)

`FuzzSetCodeTxUnmarshal` mutates the wire payload of an EIP-7702 `SetCodeTx` proto and asserts: `proto.Unmarshal` never panics; `AsEthereumData` / `Validate` never panic on a successfully-unmarshaled payload; and every `Validate` failure wraps one of the typed sentinels enumerated in `validateSentinels` (currently `ErrInvalidGasCap`, `ErrInvalidAmount`, `ErrInvalidGasFee`, `ErrSetCodeMissingTo`, `ErrSetCodeEmptyAuthList`, `ErrInvalidSigner`, `ErrInvalidChainID`, `ErrInvalidAddress`). A Validate failure that does NOT wrap one of these is treated as a sentinel leak and fails the iteration. The seed corpus hand-crafts the canonical shape plus every documented Validate failure mode (15 seeds), so the harness behaves as a normal table-driven test before any fuzzing engine engages.

#### `indexer/kv_indexer_test.go`

`TestKVIndexerSetCodeTx` and `TestKVIndexerSetCodeTxExceedsBlockGasLimit` close a gap in the custom KV indexer's coverage: the existing table-driven `TestKVIndexer` builds via `types.NewTx`, which has no `SetCodeTx` arm, so EIP-7702 envelopes never reached the indexer through the existing assertions. The new cases construct a real type-`0x04` envelope signed with `ethsecp256k1`, encode it through `MakeEncodingConfig().TxConfig`, thread the same `client.Context` into the indexer so codec round-trips line up, and pin that tx hash, height, eth-tx-index, status, gas-used, and from-address survive the round trip through the KV index — including the failed-status path when the block gas limit is exhausted.

#### `.claude/MEZO-4013-eip7702-set-code/{spec,roadmap}.md`

The spec's filename references move from the old `Eip7702MezoDivergence.test.ts` / `Eip7702*.test.ts` to the underscored `Eip7702_MezoDivergence.test.ts` / `Eip7702_*.test.ts` that actually ship. The roadmap reflects the consolidated suite layout (one spec-compliance suite plus the divergence tripwires, fuzz target, shared helpers, and ethers bump), the KV-indexer test backfill, and supersession of the `Eip7702SendRawTx` smoketest. The Prague section in `docs/evm-compatibility.md` is explicitly marked deferred to a small follow-up so prose-edit review stays separate from implementation review.

### Testing

`go test -race -timeout 5m -run TestKVIndexerSetCodeTx ./indexer/...` passes (two cases). `go test -race -timeout 5m -run FuzzSetCodeTxUnmarshal ./x/evm/types/...` passes (15 seed cases). `go vet ./indexer/... ./x/evm/types/...` is clean.

The system suite (`./tests/system/system-tests.sh Eip7702_SpecCompliance Eip7702_MezoDivergence`) runs against a localnode (`make localnode-bin-start`, chain id `0x7b7b` / `31611`, `prague_time = 0`). All 21 cases pass — 17 spec-compliance + 4 divergence tripwires.

---

### Author's checklist

### Reviewer's checklist


tomaszslabon (approved):

tomaszslabon:
The tests look good and pass.

PR #686, opened 2026-05-12
Wrap up EIP-7702 — v11.0.0 upgrade handler and docs migration

lukasz-zimnoch:
Closes: [MEZO-4013](https://linear.app/thesis-co/issue/MEZO-4013)
Depends on: https://github.com/mezo-org/mezod/pull/685

### Introduction

After this PR, mezod ships everything needed to activate EIP-7702 on the live mainnet and testnet networks, and the canonical EIP-7702 spec lives next to the rest of the chain's spec material rather than under `.claude/`. New chains already pick Prague up at genesis because `DefaultChainConfig()` writes `PragueTime = 0`; the live networks need a one-shot store mutation because the `PragueTime` field was added after those chains had already produced blocks. The remaining work to ship 7702 on those networks is the operational step of scheduling the upgrade plan through the `Upgrade` precompile.

#### New `app/upgrades/v11_0` package

A planned upgrade handler under `v11_0.Upgrade` writes the upgrade block's own unix timestamp into the stored `ChainConfig.PragueTime`. Any value strictly greater than the max historical block time keeps `IsPrague(t) = t >= *PragueTime` evaluating to `false` for every pre-upgrade block (so replay does not diverge); the block's own timestamp is a convenient pick because CometBFT already enforces strictly-monotonic block times. The upgrade block itself becomes the first block at which `IsPrague(t) = true`. The handler routes through `EvmKeeper.SetParams`, which runs `ChainConfig.Validate()` (and through it geth's `CheckConfigForkOrder`), so the handler is defended against accidentally regressing the fork-order invariant on the same path. Registered in `app/upgrades.go`.

#### `docs/spec/eip7702-set-code.md` (new, moved from `.claude/`)

The in-flight design doc graduates to the canonical `docs/spec/` location alongside `docs/spec/eth-simulate-v1.md`. The top-of-file blockquote announcing the future move is dropped. The Usage examples section is reworked to match the eth-simulate-v1.md style: a short prose paragraph per example with a self-contained JSON-RPC envelope underneath, three examples (install / call / read-designator). The install example uses the structured `eth_sendTransaction` form so the documented `authorizationList` shape is visible inline rather than collapsed into an opaque `0x04` hex blob — the implementation summary still notes that `eth_sendRawTransaction` is the consensus surface. A trailing pointer paragraph references the end-to-end localnode flows in `tests/system/test/Eip7702_*.test.ts` and the shared helpers in `tests/system/test/helpers/eip7702.ts`. The proto field name (`auth_list`) and its JSON-RPC alias (`authorizationList`) are clarified on first use so readers do not think they are two different fields.

The `.claude/MEZO-4013-eip7702-set-code/roadmap.md` companion file is deleted along with the moved spec; the directory is empty afterwards.

#### `docs/evm-compatibility.md`

A new Prague subsection is appended to the "EVM forks post-London" block, right after Cancun. It follows the per-fork structure used by Cancun and Shanghai: a one-paragraph intro framing the fork on Mezo, then per-EIP bullets with Description / Mezo implementation / Ref. Only EIP-7702 is enumerated; the other Prague EIPs (EIP-2537, 2935, 6110, 7002, 7251, 7549, 7623, 7685, 7691) are explicitly out-of-scope and can be added when individual EIPs ship. The Mezo-implementation bullet captures the four user-visible behavior points (install/rotate/clear via `0x04` transactions, EIP-3607 exemption for delegated EOAs, `EXTCODE*` returning the raw 23-byte designator, precompile-target authorizations rejected) and links to `docs/spec/eip7702-set-code.md` as the canonical behavior reference.

Four prose patches address areas the doc's pre-Prague phrasing now gets wrong:

- `eth_getCode`: the old "empty (`0x`) → EOA" heuristic is rewritten to acknowledge the third case Prague introduces — a delegated EOA whose code field carries the 23-byte `0xef0100 || target` designator rather than empty bytes.
- `eth_sendTransaction` / `eth_call` / `eth_estimateGas`: the call-object enumeration gains an `authorizationList` entry describing the per-tuple shape (`chainId`, `address`, `nonce`, `v`, `r`, `s`) and pointing at the spec. The three new entries share phrasing so they read as a uniform set.
- `eth_sendRawTransaction`: a one-sentence note that signed type-`0x04` transactions are accepted on Prague-active chains.

#### `docs/upgrades.md`

A `v11.0.0` row is added to both the testnet and mainnet historical tables with `TBD` block heights, matching the convention every prior version row used while in pre-deployment. Heights get filled in at deployment time via the `Upgrade` precompile gov flow.

### Testing

- `go build ./...` — the new package compiles and `app/upgrades.go` type-checks against the package's exported `Upgrade` value.
- `golangci-lint run ./app/upgrades/v11_0/...` — clean.
- `markdownlint` on the three touched docs — no new findings; the pre-existing `MD059` on `docs/evm-compatibility.md` line 15 (`[here]`) and the `MD060` column-style warnings on the `v10.0.0` row of `docs/upgrades.md` are unchanged.
- The handler is a single store mutation guarded by `SetParams` → `ChainConfig.Validate()` → geth's `CheckConfigForkOrder`; its live-chain effect is only observable on a chain that was started before the `PragueTime` field existed (i.e. the live networks), so verification happens at deployment time on testnet first, then mainnet. The end-to-end EIP-7702 behavior the handler unlocks is already covered by the `Eip7702_*.test.ts` system suites against localnode.

---

### Author's checklist

### Reviewer's checklist


tomaszslabon:
Verified the upgrade on localnet.
Setup:
-  localnet on commit 0b71a91c (v10.0.0), 4 nodes producing blocks.

Steps:
- Scheduled `v11.0.0` upgrade at height `65` via the Upgrade precompile.
- Nodes halted at height `65` with UPGRADE "v11.0.0" NEEDED.
- Switched to `set-code-7`, rebuilt, restarted.
- Handler ran: logged running `v11.0.0` upgrade handler; Prague time updated `pragueTime=1778662081`.
- Block production resumed.


Verification:
- mezod evm params confirmed `prague_time` set to upgrade-block timestamp; `shanghai_time` and `cancun_time` remained `0`. Fork order intact.
```
./build/mezod q evm params --node tcp://localhost:26657 --output json \
  | jq '.params.chain_config | {shanghai_time, cancun_time, prague_time}'

{
  "shanghai_time": "0",
  "cancun_time": "0",
  "prague_time": "1778662081"
}
```

tomaszslabon (approved):

PR #690, opened 2026-05-13
Tracer compatibility update

tomaszslabon:
Closes: https://linear.app/thesis-co/issue/MEZO-4011/tracer-compatibility

### Introduction

This PR updates Mezod's tracing integration after the `go-ethereum` upgrade to `v1.16`.
It aligns the tracer setup with the new geth tracing hooks and makes state-aware tracers observe the same StateDB mutations that happen during EVM execution.

### Changes

- Updated access-list tracer fork detection to use the actual block timestamp, so timestamp-activated precompiles are handled correctly.
- Excluded Mezo custom precompiles from access-list tracer output, matching their warm-by-default behavior.
- Installed EVM tracer hooks on StateDB during message execution so debug tracers can observe StateDB events.
- Emitted tracing hooks from StateDB balance, nonce, code, storage, log, and selfdestruct mutations.

### Testing

Unit tests: `go test ./x/evm/types/... ./x/evm/statedb/... ./x/evm/keeper/...`
Manually tested on local network:
- `debug_traceTransaction` with `callTracer`, `prestateTracer`, and `4byteTracer`.
- traced BTC-token custom precompile transaction.
- traced normal Solidity contract storage write.
- traced same-transaction `SELFDESTRUCT`.
- `debug_traceBlockByNumber` on a non-empty block.

---

### Author's checklist

### Reviewer's checklist


x/evm/keeper/state_transition.go
```diff
@@ -533,6 +533,10 @@ func (k *Keeper) applyMessageWithConfig(
 	}
 
 	evm := k.NewEVMWithOverrides(ctx, msg, cfg, tracer, stateDB, evmOverrides)
+	if evm.Config.Tracer != nil {
+		stateDB.SetTracingHooks(evm.Config.Tracer)
+		defer stateDB.SetTracingHooks(nil)
+	}
```
lukasz-zimnoch: Note that this effectively nuke the already-set hooks. This is the case for simulate V1 which sets their own hooks before we reach this point here. I think the cleanest path is to modify stateDB to handle multiple hooks. We can do this though a mapping so each hook has a key that can be later used for de-registration.

Note: I was initially thinking about using the wrapper pattern geth uses (https://github.com/mezo-org/go-ethereum/blob/master/core/state/statedb_hooked.go) but this will be more complex as our stateDB implementation is a superset of geth's and we cannot easily replace current pointer references with the geth `StateDB` interface. Migrating this code seems to be an overkill now given we just want to add tracing hooks.

x/evm/statedb/statedb.go
lukasz-zimnoch, line 1: Let's ensure we call the hooks on the same methods geth does: https://github.com/mezo-org/go-ethereum/blob/master/core/state/statedb_hooked.go

lukasz-zimnoch, line 1: I don't quite like the fact we add inline tracer logic within critical stateDB methods. Can we at least extract tracer logic into helper functions? Ideally, each function should have a one-line addition calling the tracer helper. This will be also helpful in the context of multi-hook model proposed in https://github.com/mezo-org/mezod/pull/690/changes#r3241133920

x/evm/statedb/statedb.go
```diff
@@ -716,19 +737,36 @@ func (s *StateDB) RegisterCachedCtxCheckpoint(addr common.Address, cachedCtxChec
...
+		prevHash := crypto.Keccak256Hash(prev)
+		codeHash := crypto.Keccak256Hash(code)
 		stateObject.SetCode(crypto.Keccak256Hash(code), code)
+		if prevHash != codeHash && s.tracingHooks != nil {
+			if s.tracingHooks.OnCodeChangeV2 != nil {
+				s.tracingHooks.OnCodeChangeV2(addr, prevHash, prev, codeHash, code, reason)
+			} else if s.tracingHooks.OnCodeChange != nil {
+				s.tracingHooks.OnCodeChange(addr, prevHash, prev, codeHash, code)
+			}
+		}
```
lukasz-zimnoch: We are doing redundant `crypto.Keccak256Hash` here. First `code` can be hashed once and re-used. Second, `prevHash` can be computed only when hooks are non-nil.

x/evm/keeper/keeper.go
```diff
@@ -295,7 +296,15 @@ func (k *Keeper) PostTxProcessing(ctx sdk.Context, msg core.Message, receipt *et
 
 // Tracer return a default vm.Tracer based on current keeper state
 func (k Keeper) Tracer(ctx sdk.Context, msg core.Message, ethCfg *params.ChainConfig) *tracers.Tracer {
-	return types.NewTracer(k.tracer, msg, ethCfg, ctx.BlockHeight())
+	customPrecompiles := maps.Keys(k.resolveCustomPrecompiles(ctx))
```
lukasz-zimnoch: This is relevant only for `TracerAccessList`. Maybe we can use lazy-evaluation here and pass it as a function that will be called only when the tracer is `TracerAccessList`?

x/evm/keeper/state_transition.go
```diff
@@ -533,6 +533,10 @@ func (k *Keeper) applyMessageWithConfig(
 	}
 
 	evm := k.NewEVMWithOverrides(ctx, msg, cfg, tracer, stateDB, evmOverrides)
+	if evm.Config.Tracer != nil {
```
lukasz-zimnoch: Note that this is actually an always-true condition as we use `NoOpTracer` by default. That said, all the conditional tracing branches of stateDB we added in this PR will fire just to hit the no-op hook methods of `NoopTracer`. This is suboptimal and may lead to resource waste. I would either don't set stateDB hooks if we detect `NoopTracer` or modify `NoOpTracer` to:

```
func NewNoopTracer() (*tracers.Tracer, error) {
	t := &NoOpTracer{}
	return &tracers.Tracer{
		Hooks: new(tracing.Hooks),
		GetResult: t.GetResult,
		Stop:      t.Stop,
	}, nil
}

// remove all unused methods of `NoOpTracer`
```
so their all methods in its `Hooks` field are nil. This however, may not fully solve the problem as stateDB paths that do `s.tracingHooks != nil` will still go through.

x/evm/keeper/keeper.go
tomaszslabon: Done in d4a82d81f0212ea1a091692244cd78c6674a2861.

x/evm/statedb/statedb.go
tomaszslabon: Done in 9af2589827831fcf86513d5fe62393d798f67545.

x/evm/statedb/statedb.go
tomaszslabon: Extracted tracing into helpers in 75930b6ac6c9c7f0b64cba9ee41ed00fb04ca92f.

x/evm/keeper/state_transition.go
tomaszslabon: Done in 5330377e29965d050faec3da861e2fc5e4730a91.

x/evm/keeper/state_transition.go
tomaszslabon: This is probably outdated after the recent changes for other review comments. We now install `StateDB` hooks only when the tracer actually needs them. In `eth_simulateV1`, the simulate tracer is passed into applyMessageWithConfig, so hooks are still active during execution. I’d avoid adding multi-hook support now. WDYT?

x/evm/statedb/statedb.go
tomaszslabon: We now call hooks from the same mutation methods as geth:
- AddBalance / SubBalance
- SetNonce
- SetCode
- SetState
- AddLog
- SelfDestruct

x/evm/statedb/statedb.go
```diff
@@ -356,19 +424,24 @@ func (s *StateDB) SelfDestruct(addr common.Address) uint256.Int {
...
-		n    = new(uint256.Int)
+		prev         = new(uint256.Int).Set(stateObject.Balance())
+		n            = new(uint256.Int)
+		prevCode     []byte
+		prevCodeHash common.Hash
 	)
+	if s.tracingHooks != nil && (s.tracingHooks.OnCodeChangeV2 != nil || s.tracingHooks.OnCodeChange != nil) {
+		prevCode = append([]byte(nil), stateObject.Code()...)
+		prevCodeHash = crypto.Keccak256Hash(prevCode)
+	}
```
lukasz-zimnoch: Let's also fold this into `traceSelfDestructCodeChange`. We can set `prevCode` without any condition here inside `SelfDestruct` (like we do for `prev` balance) and compute `prevCodeHash` inside  `traceSelfDestructCodeChange`.

x/evm/keeper/state_transition.go
```diff
@@ -68,6 +68,19 @@ type EVMOverrides struct {
...
+	if hooks == nil {
+		return false
+	}
+	return hooks.OnBalanceChange != nil ||
+		hooks.OnNonceChange != nil ||
+		hooks.OnNonceChangeV2 != nil ||
+		hooks.OnCodeChange != nil ||
+		hooks.OnCodeChangeV2 != nil ||
+		hooks.OnStorageChange != nil ||
+		hooks.OnLog != nil
```
lukasz-zimnoch: Shouldn't we rather name this function `hasTracingHooks` and nil-check all fields of `tracing.Hooks`? This way, we don't couple with internal state DB implementation and use a simple rule: if there are tracing hooks (at least one method is non-nil) we set them in the state DB and that's it.

Moreover, worth dropping a comment why we check all the fields and that we need to re-visit this site on each geth upgrade as `tracing.Hooks` may gain/lose fields over time.

x/evm/keeper/state_transition.go
lukasz-zimnoch: I don't think so. Even if this is not an issue right now for the simulate v1 scenario, the entire mechanism is flaky and will surely bite us in other conditions. We need to either add multi-hook support or guarantee only one hook can be set at a time. Silent overwrite is not a proper solution here.

x/evm/statedb/statedb.go
tomaszslabon: Done in 15c21b7885443f3b5cf590f6492a5457ca84b2ab.

x/evm/keeper/state_transition.go
tomaszslabon: Done in 81ac1eddaf605d97b3511397f1c3a2d9212c7dfb.

x/evm/keeper/state_transition.go
tomaszslabon: @lukasz-zimnoch Added multi-hook support to `StateDB` in 47a1b22ed7ae853c6ac811e57bbbfafafdc60003.
Please, double check if it's ok.

x/evm/statedb/statedb.go
```diff
@@ -157,10 +157,115 @@ func (s *StateDB) GetContext() sdk.Context {
...
+	for _, hooks := range s.tracingHooks {
+		if hooks.OnCodeChangeV2 != nil {
+			hooks.OnCodeChangeV2(addr, prevHash, prev, codeHash, code, reason)
+		} else if hooks.OnCodeChange != nil {
+			hooks.OnCodeChange(addr, prevHash, prev, codeHash, code)
+		}
+	}
+}
+
+func (s *StateDB) traceSelfDestructCodeChange(addr common.Address, stateObject *stateObject) {
```
lukasz-zimnoch: Can't we just pass `code` as argument instead of the entire `stateObject`? Passing an entire object to a function that doesn't need it is a strong code smell -  especially when we consider how critical `stateObject` is.

x/evm/statedb/statedb.go
```diff
@@ -157,10 +157,115 @@ func (s *StateDB) GetContext() sdk.Context {
...
+}
+
+func (s *StateDB) traceCodeChange(
+	addr common.Address,
+	prev []byte,
+	code []byte,
+	codeHash common.Hash,
+	reason tracing.CodeChangeReason,
+) {
+	prevHash := crypto.Keccak256Hash(prev)
```
lukasz-zimnoch: We can early return when `len(s.tracingHooks) == 0 ` to avoid this keccak when there is no tracers.

x/evm/statedb/statedb.go
lukasz-zimnoch, line 1: WDYT about extracting all tracing functions (`AddTracingHooks` and all the `trace*` ones) to a separate file `statedb_tracing.go` (same for tests).

x/evm/statedb/statedb.go
tomaszslabon: Done in 3e33ecf7818235e2927ae1dbc30d564079d8f447.

x/evm/statedb/statedb.go
tomaszslabon: Done in d2a6bb4fe28206dca420339d2b12a5f39023c495.

x/evm/statedb/statedb.go
tomaszslabon: Done in 839fdb148520dad73a542455208ccadcaff4cb35.

lukasz-zimnoch (approved):

PR #694, opened 2026-05-15
Document Prague and Osaka EIP coverage in evm-compatibility.md

lukasz-zimnoch:
Closes: [MEZO-4411](https://linear.app/thesis-co/issue/MEZO-4411/update-evm-compatibilitymd)
References: [MEZO-4412](https://linear.app/thesis-co/issue/MEZO-4412/final-completness-check)

### Introduction

`docs/evm-compatibility.md` is Mezo's public reference for which EVM forks and EIPs the chain implements, where it diverges from upstream, and how those choices manifest at the JSON-RPC surface. Until now it covered Paris, Shanghai, and Cancun in full, but only EIP-7702 from the Prague set — the rest of Prague and all of Osaka were undocumented. This PR fills that gap so external integrators (wallets, tracers, indexers, dApp authors) can see Mezo's stance on every Prague and Osaka execution-layer EIP without reading the source tree.

#### `docs/evm-compatibility.md`

The "Prague" subsection is expanded from a one-EIP placeholder into a full Pectra execution-layer summary. The existing EIP-7702 entry is retained verbatim; new entries cover EIP-2537, EIP-2935, EIP-6110, EIP-7002, EIP-7623, EIP-7685, EIP-7840, and EIP-7642. The subsection opens with how Prague is activated on Mezo (v11.0 upgrade on living chains, genesis on fresh chains) and points at EIP-7600 as the meta-EIP reference, mirroring the Cancun subsection's pointer to EIP-7569.

A new "Osaka (Fusaka)" subsection is added in the same format, covering EIP-7823, EIP-7825, EIP-7883, EIP-7934, EIP-7935, EIP-7939, EIP-7951, EIP-7892, EIP-7918, EIP-7642, and EIP-7910. The opening paragraph records that Osaka is activated on Mezo alongside Prague — the v11.0 upgrade flips both fork times together, so Cancun → Prague → Osaka collapse into a single chain upgrade.

Each EIP entry uses the existing `- EIP-N (name) / Description / Mezo implementation / Ref` shape, with the "Mezo implementation" body kept to the integrator-visible facts: status (supported / not applicable / not supported / out of scope), addresses callers interact with, observable RPC behavior, and any practical Mezo-specific divergence. Internal source paths, helper names, and "how it's wired" detail are intentionally omitted.

Supported (work as specified by the EIP): EIP-2537, EIP-7623, EIP-7702 (already documented), EIP-7823, EIP-7883, EIP-7939, EIP-7951.

[...]

PR #695, opened 2026-05-18
Activate the Osaka fork at genesis and in the v11 upgrade handler

lukasz-zimnoch:
References: [MEZO-4412](https://linear.app/thesis-co/issue/MEZO-4412/final-completness-check)

### Introduction

Activates the Osaka (Fusaka) EVM fork on Mezo. New chains start with Osaka enabled at genesis; living chains pick it up when the `v11.0.0` planned upgrade runs.

Mezo collapses Cancun → Prague → Osaka into a single chain halt (Cancun → Prague was never rolled out separately), so the same upgrade block flips both `PragueTime` and `OsakaTime`.

#### Genesis defaults

`x/evm/types.DefaultChainConfig` now defaults `OsakaTime` to zero alongside the earlier fork times. The Osaka-related geth changes (EIP-7823 MODEXP input bound, EIP-7825 per-tx gas cap, EIP-7883 MODEXP gas schedule, EIP-7939 `CLZ`, EIP-7951 secp256r1 precompile) already ride through Mezo's EVM, and `params.DefaultBlobSchedule` covers Cancun/Prague/Osaka — so the fork is safe to activate at genesis.

#### Living-chain activation in v11

The `v11.0.0` upgrade handler previously set only `PragueTime` to the upgrade block's timestamp. It now also flips `OsakaTime` to the same timestamp via a separate `setOsakaTime` helper that mirrors the existing `setPragueTime` (own log breadcrumbs, own error wrapping at the call site). Both must move forward together at the upgrade block — otherwise `CheckConfigForkOrder` rejects the post-upgrade config or the chain ends up on the wrong fork.

#### Test-setup fallout

Ante and EVM-keeper test setups that pin the chain to pre-London for fee-market scenarios already set every later fork time to `math.MaxInt64` to keep the disabled-fork ordering valid. They now do the same for `OsakaTime`; without this, the new genesis default flips Osaka to zero while Prague is `MaxInt64` and `CheckConfigForkOrder` rejects the genesis.

### Testing

- `go build ./...` clean.
- `go test ./x/evm/... ./app/ante/... ./app/upgrades/... ./app/...` — all green, including the chain-config validation table and the cross-wiring sentinel test in `x/evm/types`.
- Manual sanity: `ChainConfig.Validate()` accepts the new default (covered by the existing `TestChainConfigValidate` "default" case in `chain_config_test.go`).

---

### Author's checklist

### Reviewer's checklist


tomaszslabon:
@lukasz-zimnoch 
This PR activates the Osaka fork.
One of the features introduced in Osaka is `EIP-7825`, which imposes some cap on how much gas a single transaction can consume (see how it's [enforced in geth](https://github.com/mezo-org/go-ethereum/blob/v1.16.9-mezo0/core/state_transition.go#L327-L331)). If we want mezod to follow the same logic, we would have to add a similar check. Perhaps we could add it in [here in applyMessageWithConfig](https://github.com/mezo-org/mezod/blob/osaka-fork-activation/x/evm/keeper/state_transition.go#L565). The check could look like this:
```
if !msg.SkipTransactionChecks && rules.IsOsaka && msg.GasLimit > params.MaxTxGas {
	return nil, nil, errorsmod.Wrapf(
		core.ErrGasLimitTooHigh,
		"apply message: cap %d, tx %d",
		params.MaxTxGas,
		msg.GasLimit,
	)
}
```

lukasz-zimnoch:
> @lukasz-zimnoch This PR activates the Osaka fork. One of the features introduced in Osaka is `EIP-7825`, which imposes some cap on how much gas a single transaction can consume (see how it's [enforced in geth](https://github.com/mezo-org/go-ethereum/blob/v1.16.9-mezo0/core/state_transition.go#L327-L331)). If we want mezod to follow the same logic, we would have to add a similar check. Perhaps we could add it in [here in applyMessageWithConfig](https://github.com/mezo-org/mezod/blob/osaka-fork-activation/x/evm/keeper/state_transition.go#L565). The check could look like this:
> 
> ```
> if !msg.SkipTransactionChecks && rules.IsOsaka && msg.GasLimit > params.MaxTxGas {
> 	return nil, nil, errorsmod.Wrapf(
> 		core.ErrGasLimitTooHigh,
> 		"apply message: cap %d, tx %d",
> 		params.MaxTxGas,
> 		msg.GasLimit,
> 	)
> }
> ```

Please see the WIP: https://github.com/mezo-org/mezod/pull/694 that clarifies this EIP among others.

tomaszslabon (approved):

PR #701, opened 2026-05-29
RFC-7: Mezo bridge intents

lukasz-zimnoch:
Closes: [MEZO-4500](https://linear.app/thesis-co/issue/MEZO-4500)

### Introduction

The Mezo bridge today moves BTC and ERC20 assets between Ethereum and Mezo as plain value transfers — the user picks an asset, an amount, and a recipient, the bridge mints or releases on the other side, and the flow ends there. A growing class of cross-chain use cases needs more than that: a Mezo user may want to deposit BTC as collateral on an Ethereum lending market and receive the borrowed stablecoin back on Mezo in a single user-facing action, without ever custodying funds on Ethereum. Today those flows require either a custom bridge integration per protocol or a series of manual steps from the user. This PR adds RFC-7, a design proposal for an *intent* layer on top of the existing bridge so a single user action can declaratively express "bridge these assets and do X with them on the other side". The design follows the shape of [ERC-7683](https://eips.ethereum.org/EIPS/eip-7683) so generic, permissionless third-party fill infrastructure can plug in without protocol-specific code. This is a documentation-only proposal; no protocol code ships in this PR.

#### `docs/rfc/rfc-7.md`

A new RFC covering the symmetric Mezo↔Ethereum intent design:

- Bridge-with-payload entry points on `MezoBridge` that carry an opaque payload alongside a transfer, plus a companion `PayloadRecorded` event for observability.
- A registry-driven bridge-hook mechanism that fires on allowlisted recipients after each transfer completes, kept orthogonal to the intent layer.
- Per-protocol Resolvers that act as recipient, hook target, ERC-7683 `resolve()` surface, custodian, and executor for their own Orders, with no central settler or shared custodian.
- Untrusted, permissionless Solvers that watch bridge events and submit fills, plus a worked remote-borrow example showing how a generic Solver discovers and fills an Order without protocol-specific knowledge.
- The rollout shape on a live chain: a coordinated halt via the Upgrade primitive paired with the `MezoBridge` proxy upgrade, and additive `WithPayload` events.
- Open considerations left for later: optimistic fill, gasless bridge-with-payload, a public discovery layer, and Morpho V2 cross-chain markets.

### Testing

Documentation-only change — no executable code ships in this PR. `markdownlint docs/rfc/rfc-7.md --config .markdownlint.yml` passes.

---

### Author's checklist

### Reviewer's checklist

### DIFF
--- a/docs/rfc/rfc-7.md
+++ b/docs/rfc/rfc-7.md
@@ -199,7 +199,7 @@ This keeps the on-chain data and the consensus pipeline bounded to a fixed
 Both origin-chain contracts emit the same direction-agnostic event:
 
 ```solidity
-event PayloadRecorded(bytes32 indexed payloadHash, bytes payload);
+event PayloadRecorded(uint256 indexed sequence, bytes32 indexed payloadHash, bytes payload);
 ```
 
 It is emitted in the same transaction as the main bridge event.
@@ -213,15 +213,16 @@ happened. A Solver therefore:
 
 1. Detects a bridge-with-payload finalization on the destination chain (the
    Order is ready to fill).
-2. Looks up the matching `PayloadRecorded(payloadHash, payload)` on the
-   origin chain to get the plain payload bytes.
+2. Looks up the matching `PayloadRecorded` on the origin chain by `sequence`
+   (the unique key it already has from step 1) to get the plain payload bytes.
 
 `payloadHash` is **not** guaranteed unique across pending entries by the
-bridge - two users can submit the same payload, and the companion event is
-keyed on `payloadHash` for indexed lookup only. The bridge's monotonically
+bridge - two users can submit the same payload. The bridge's monotonically
 assigned `sequence` number is therefore the only *bridge-guaranteed* unique
-key. A Resolver that prefers to address Orders by `payloadHash` instead must
-enforce payload uniqueness itself and push that responsibility onto the user.
+key, so the companion event indexes it alongside `payloadHash` and Solvers
+correlate by `sequence`. A Resolver that prefers to address Orders by
+`payloadHash` instead must enforce payload uniqueness itself and push that
+responsibility onto the user.
 
 #### Bridge entry schema
 
@@ -487,10 +488,12 @@ protocol, or that the Order is a "borrow". It follows ERC-7683 mechanically:
 1. **Discover an Order.** The Solver watches bridge events. When
    `MezoBridge.AssetsUnlockWithPayloadConfirmed(…, payloadHash)` fires with a
    non-zero `payloadHash`, an Order exists. The Solver fetches the plain
-   `payload` from the matching `PayloadRecorded(payloadHash, payload)` event
-   emitted on Mezo by the `AssetsBridge` precompile (where the user originally
-   called `bridgeOutWithPayload`) and verifies
-   `keccak256(payload) == payloadHash`. That hash is the Order's key.
+   `payload` from the matching `PayloadRecorded` event emitted on Mezo by the
+   `AssetsBridge` precompile (where the user originally called
+   `bridgeOutWithPayload`) and verifies `keccak256(payload) == payloadHash`.
+   That hash is the Order's key *here* because this Resolver enforces payload
+   uniqueness; the generic discovery path keys by the unique `sequence`
+   instead.
 2. **Resolve.** The Solver calls `Resolver.resolve(payload)` via `eth_call`.
    It receives a `ResolvedOrder` describing Steps (calls to make), Variables
    (placeholders — here just the Solver's own payout address), Payments (what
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-7.md
```diff
@@ -0,0 +1,725 @@
...
+
+#### What the user does
+
+The user makes a single call on Mezo:
+
+```solidity
+AssetsBridge.bridgeOutWithPayload(
+    BTC,            // token
+    btcAmount,      // amount
+    ETHEREUM,       // chain
```
tomaszslabon, line 458: Do we need the target chain argument? 
`bridgeOutWithPayload` will only work when bridging to Ethereum as the provided `recepient` (resolver) must be a smart contract. Unless we want to keep the target chain for the sake of completeness (ordinary `bridgeOut` has target chain) and revert if Bitcoin is provided here.

docs/rfc/rfc-7.md
```diff
@@ -0,0 +1,725 @@
...
+
+        // CEI. `admitted[payloadHash]` stays set, so this payload can never be
+        // re-admitted once the entry is cleared.
+        delete entries[payloadHash];
+
+        IERC20(TBTC).approve(AAVE_POOL, e.amount);
+        IPool(AAVE_POOL).supply(TBTC, e.amount, address(this), 0);
+        IPool(AAVE_POOL).borrow(USDC, p.usdcToBorrow, 2, 0, address(this));
+
+        IERC20(USDC).transfer(msg.sender, p.solverFee);
```
tomaszslabon: Nitpick: this should probably be 
```suggestion
        IERC20(USDC).transfer(solverPayoutAddress, p.solverFee);
```
where `solverPayoutAddress` is the address that gets the solver's payment (fee for calling `fill()`). It should be passed as an argument to `fill()`. 
The `resolve()` function lets the solver choose the payment recipient:
```
bytes memory varRecipient = abi.encodeWithSignature("PaymentRecipient()");
```

docs/rfc/rfc-7.md
```diff
@@ -0,0 +1,725 @@
...
+is observed on the destination chain. Exactly how that finalization surfaces
+differs by direction; for discovery a Solver only needs to detect that it has
+happened. A Solver therefore:
+
+1. Detects a bridge-with-payload finalization on the destination chain (the
+   Order is ready to fill).
+2. Looks up the matching `PayloadRecorded(payloadHash, payload)` on the
+   origin chain to get the plain payload bytes.
+
+`payloadHash` is **not** guaranteed unique across pending entries by the
```
tomaszslabon, line 221: We probably should add the sequence number when emitting `PayloadRecorded` events:
```
PayloadRecorded(uint256 indexed sequence, bytes32 indexed payloadHash, bytes payload)
```

pdyraga:
I haven't had a chance to look into the proposal yet, but flagging that we should look at Morpho Midnight and ensure the approach is compatible. Midnight is architecturally an intent protocol already - the offer happens off-chain with on-chain settlement, separating the expression of intent from sourcing capital. We can't use the existing Morpho Blue as-is for remote-sourcing of capital - the liquidity needs to be there. Morpho Midnight solves the liquidity fragmentation within a single chain but it's a very similar async pattern to what we are trying to build to source the capital from the outside.

docs/rfc/rfc-7.md
lukasz-zimnoch: Kept the chain argument for parity with bridgeOut and made it revert for any non-Ethereum target. f74d4f40

docs/rfc/rfc-7.md
lukasz-zimnoch: fill now takes the payout address and pays it, and resolve threads variable 0 (PaymentRecipient) into the step. 1a39b452

docs/rfc/rfc-7.md
lukasz-zimnoch: Added sequence as an indexed field and re-keyed the generic discovery steps on it. The worked example stays on payloadHash since it enforces uniqueness itself. e420d78b

lukasz-zimnoch:
I dug into Midnight a bit, and I don't think it competes with what RFC-7 does - they sit at different levels. RFC-7 is really just a transport intent: bridge these assets and do something with them on the other side. It has no idea what lending is and doesn't care where the liquidity comes from. Midnight is the lending intent itself - off-chain offers that don't tie up capital, settled on-chain, with liquidity pulled in only at settlement. So Midnight fits *inside* an RFC-7 intent rather than sitting next to it.

The pieces line up well. A Midnight-backed Resolver receives the bridged collateral, the Solver plays the Midnight taker - finding lender offers off-chain and handing them to the Resolver to settle the loan - and then the borrowed funds bridge back. Those offers are just the kind of off-chain input a Solver already has to go discover, so there's nothing new on the bridge side; Midnight is one more Resolver. The only real shift is that the user expresses terms (a max rate, a term) instead of a fixed borrow amount, since the rate falls out of whatever offers are around at settlement. And the "liquidity has to be there" worry mostly goes away - RFC-7 sources it at fill time, which is exactly how Midnight works.

PR #711, opened 2026-07-08
Add incident report for July 2026 LevelDB corruption

lukasz-zimnoch:
### Introduction

Adds an incident report for the July 2026 LevelDB corruption: on 2026-07-02 the testnet archive node started crash-looping with a corrupted database after an OOM-kill wave, and in the following days mainnet operators hit the same corruption while restarting their nodes for v11.0.1. The report exists both as an internal record and as a document to share with affected node operators, so it is written for readers without deep mezod or LevelDB knowledge.

#### New report `docs/reports/2026-07-02-leveldb-corruption.md`

Follows the structure of the existing `2025-03-12-matsnet-halt.md` report, extended with operator-facing sections:

- Summary: how unclean shutdowns (SIGKILL, OOM kill) corrupt LevelDB's `MANIFEST`, and the two paths that led there (testnet OOM kills; mainnet hot restarts for v11.0.1 under pre-12.1.0 validator-kit default stop timeouts).
- Symptoms: the exact startup panic, the databases that can be affected (including `evmindexer.db` on nodes with the EVM indexer enabled), and exit code 137 as the early-warning signal.
- Recovery options as step-by-step guides, safest first: restore from backup (own backup, or a third-party data-directory archive from services like Nodes Hub / Imperator with a bold non-endorsement disclaimer), re-sync via CometBFT state sync (automatic snapshot download from peers; Mezo team serves state sync snapshots only on testnet), and the experimental offline repair used on the archive node (explicitly marked dangerous, last resort). All guides cover preserving `priv_validator_state.json` to avoid double-signing, and warn against state-syncing an archive node.
- Prevention: raise stop timeouts to 120s per deployment method (shipped in validator-kit 12.1.0 for those who use it), back up the data directory, alert on unclean kills.
- Impact.
- Appendix (collapsed by default): the complete single-file rebuild tool (recover / stage1 / flushruns / stage2) used to repair the archive node, with build instructions, so the repair procedure is fully self-contained and reproducible.

### Testing

Docs-only change. `markdownlint` (repo config) passes for the new file. Facts cross-checked against the incident investigation: OOM-kill event counts and timestamps from GCP Cloud Logging, shutdown-duration measurements on live nodes, and the recovery numbers from the actual archive-node rebuild. The state-sync steps match the current validator-kit runbook, `mezod tendermint unsafe-reset-all` was verified against `server/util.go`, and the appendix program was extracted from the report and compiled against the goleveldb version pinned in mezod's `go.mod`.

### DIFF
--- a/docs/reports/2026-07-02-leveldb-corruption.md
+++ b/docs/reports/2026-07-02-leveldb-corruption.md
@@ -96,6 +96,17 @@ fastest.
 3. Start `mezod`. It block-syncs the gap between the backup and the chain tip; watch the
    logs for steadily increasing block heights.
 
+If you have no backup of your own, you can restore from someone else's: third-party
+snapshot services publish downloadable archives of the data directories of their own
+nodes. Restoring works the same way, except you replace your `data/` directory with the
+content of the archive, and you are trusting a stranger's copy of the chain data. Known
+services for Mezo mainnet are
+[Nodes Hub](https://services.nodeshub.online/mainnet/mezo/snapshot) and
+[Imperator](https://www.imperator.co/services/chain-services/mainnets/mezo).
+**These services are not vetted or endorsed by the Mezo development team; use them at
+your own risk.** As with your own backup, put your saved `priv_validator_state.json`
+back after unpacking the archive.
+
 ### 2. Re-sync using state sync
 
 If you have no backup and your node is **not** an archive node, wipe the local state and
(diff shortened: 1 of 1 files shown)


docs/reports/2026-07-02-leveldb-corruption.md
```diff
@@ -0,0 +1,603 @@
...
+
+A node with a corrupted LevelDB database crash-loops on startup. The log ends with an
+error like:
+
+```text
+panic: ... leveldb: manifest corrupted (field 'comparer'): missing [file=MANIFEST-3772245]
+```
+
+The file name and the field vary. Any of the six databases in the data directory can be
+affected: `application.db`, `blockstore.db`, `state.db`, `tx_index.db`, `evidence.db` and
```
tomaszslabon: Probably also [evmindexer.db](https://github.com/mezo-org/mezod/blob/b36644a5befb9f1d404dc9d55af92d7605116ac4/server/start.go#L640-L643) could be added to the list of vulnerable databases.

lukasz-zimnoch: Added in 6054706f, with a note that it only exists on nodes running with `enable-indexer = true`.

tomaszslabon (approved):

PR #712, opened 2026-07-10
Cut Pyth price-pusher update cost on Mezo

dimpar:
## Summary

- Lowers the on-chain cost of the Mezo Pyth scheduler (price pusher) by ~5x
- The tradeoff is that a contract that liquidates or prices collateral off these feeds now need to tolerate up to a 2h-stale price in quiet markets, plus the 3% deviation band.

## Changes

- `infrastructure/kubernetes/common/pyth-scheduler-mezo/configmap.yaml`
  - `price_deviation` raised to **3%** on all feeds (fewer trigger events).
  - **`early_update` block added to every non-driver feed**, restricting it to join a batch only when it has itself moved ≥3% — suppresses the default bundling of feeds that didn't move (the biggest, most reliable lever, since ~80% of cost is per-feed).
  - `time_difference` raised to **7200** (2h staleness heartbeat) on all feeds.
- `infrastructure/kubernetes/common/pyth-scheduler-mezo/deployment.yaml`
  - Added `--pushing-frequency 30` as a burst cap so rapid BTC moves can't push more than every 30s.

### DIFF
--- a/infrastructure/kubernetes/common/pyth-scheduler-mezo/configmap.yaml
+++ b/infrastructure/kubernetes/common/pyth-scheduler-mezo/configmap.yaml
@@ -6,73 +6,100 @@ metadata:
     app: pyth-scheduler
 data:
   price-config.yaml: |
+    # --- BTC cluster: the push DRIVERS. Deviation 3%. ---
     - alias: SolvBTC/USD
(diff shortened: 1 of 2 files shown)


dimpar:
`pyth-scheduler` was replaced by `pyth-pusher` process

PR #725, opened 2026-08-11
Add checked balance synchronization for precompiles

tomaszslabon:
### Introduction

This PR adds checked synchronization for native balance changes made by precompiles.

### Changes

- Evaluates journal balance changes in execution order.
- Applies journal entries together after successful evaluation.
- Adds unit tests for balance synchronization.
- Updates the bridge-out test setup.

### Testing

```
go test ./precompile/...
```

---

### Author's checklist

### Reviewer's checklist


precompile/contract.go
```diff
@@ -280,6 +286,53 @@ func (c *Contract) syncJournalEntries(journal *StateDBJournal, stateDB *statedb.
 	}
 
 	journal.entries = nil
+	return nil
+}
+
+func (c *Contract) validateJournalBalances(journal *StateDBJournal, stateDB *statedb.StateDB) error {
```
lukasz-zimnoch: We can make this change smaller. The two-pass validate-then-apply structure is not necessary. When `Run` returns an error, the EVM reverts all state changes to the pre-call snapshot. The checkpoint comment at the top of `Run` documents this behavior. All error paths after `CommitCacheContext` already depend on this revert.

Check and apply each entry in one loop:

```go
func (c *Contract) syncJournalEntries(journal *StateDBJournal, stateDB *statedb.StateDB) error {
	for _, v := range journal.entries {
		balance := stateDB.GetBalance(v.Address)
		if v.isSub {
			if balance.Lt(v.Amount) {
				return fmt.Errorf(
					"balance underflow for address %s: balance %s, delta %s",
					v.Address.Hex(), balance, v.Amount,
				)
			}
			stateDB.SubBalance(v.Address, v.Amount, v.TracingReason)
			continue
		}

		if _, overflow := new(uint256.Int).AddOverflow(balance, v.Amount); overflow {
			return fmt.Errorf(
				"balance overflow for address %s: balance %s, delta %s",
				v.Address.Hex(), balance, v.Amount,
			)
		}
		stateDB.AddBalance(v.Address, v.Amount, v.TracingReason)
	}

	journal.entries = nil
	return nil
}
```

This removes `validateJournalBalances` and the balances map. Each check reads the live balance. Thus entries that touch the same address stay correct without a map. A failed entry causes a full revert through the snapshot. No partial application is visible.

tomaszslabon: Done in 4ec29fa83db883d47ab38b6eee9edf51c49f5b59.

precompile/contract_test.go
```diff
@@ -298,6 +300,105 @@ func TestContract_Run(t *testing.T) {
...
+	journal.SubBalance(from, uint256.NewInt(5), tracing.BalanceChangeUnspecified)
+
+	err := new(Contract).syncJournalEntries(journal, stateDB)
+
+	require.ErrorContains(err, "balance underflow")
+	require.Equal(uint256.NewInt(4), stateDB.GetBalance(from))
+	require.Len(journal.entries, 2)
+}
+
+func TestContract_SyncJournalEntries_AdditionOverflow(t *testing.T) {
```
lukasz-zimnoch, line 362: Can we also cover cumulative overflow?

tomaszslabon: Done in 871359364bd55f8c7c954004f6cfd0aa77da6819.

lukasz-zimnoch (approved):

PR #731, opened 2026-08-18
Set Prometheus scrape cadences by target cost

nkuba:
References: 

### Introduction

Both Prometheus configs set a global `scrape_interval: 2s` and no job overrode it. With seven targets across six jobs that is **302,400 scrapes/day per environment**, and it was noticed because Cloudflare showed ~41,000/day hitting `earn-bot.mezo.org` alone — close to the 43,200 a 2s interval predicts.

Nothing needed that cadence. **No job in either config has underlying data that refreshes faster than 30s**, and several targets do real work per scrape:

- `earn-bot` performs four on-chain `activePeriod` reads plus ~6 Cloudflare KV reads per scrape — ~173k RPC calls/day at 2s.
- `balance-exporter` caches nothing; both containers query the chain on every scrape — ~389k RPC calls/day at 2s.
- `blackbox` performs a full page load of `mezo.org` per scrape.
- `metrics-scraper` and `bridge-worker` serve cached gauges, refreshed by their own 30s and 1m tickers.

The 2s interval also broke the scrapes it was meant to make frequent. Prometheus caps `scrape_timeout` at `scrape_interval`, so every target had **two seconds** to respond. Slow scrapes were recorded as failures rather than slow successes, so targets flapped `up`/`down` and dashboards gapped. This is the likely source of the ~2k/day shortfall between the predicted 43.2k and the observed 41k on earn-bot. It also left `blackbox` unable to honour its own `http_2xx` module timeout of 5s.

### Changes

Cadence is set per job by what a scrape actually costs, in three tiers. The global is a fallback for jobs added later rather than a value any current job depends on, so a job added without a tier gets 30s rather than the old 2s.

| Tier | Jobs | Interval |
|---|---|---|
| Application metrics | `metrics-scraper` (15s), `bridge-worker` (30s) | 30s tier |
| API exporters | `api-metrics`, `earn-bot`, `blackbox` | 60s |
| Wallet exporters | `balance-exporter` | 5m |
| *(global fallback)* | — | 30s |

`scrape_timeout` is set once globally at 30s rather than per job. `evaluation_interval` is 15s — it governs rule evaluation rather than scraping, and staging's `PCVDebtToPayZero` uses a `for: 1m` window that wants several evaluations inside it.

#### Why `metrics-scraper` is 15s rather than the 30s tier value

Its node gauges are refreshed on a 30s ticker of its own (`flagNodePollRateDefault`; the deployment only overrides `--bridge-poll-rate`). Scraping at the same period as the exporter updates makes every scrape load-bearing — the two clocks drift against each other, and one failed scrape drops a `latest_block` datapoint and leaves a visible gap. 15s captures each update roughly twice, so a lost scrape is invisible. `bridge-worker` takes the tier value: its gauge is written on 1m loops, so 30s samples it twice per update.

#### Why one global timeout is enough

Prometheus validates an *explicit* timeout against an *explicit* interval on the same job and rejects the config when it is larger — which makes per-job timeouts look mandatory. An *inherited* timeout behaves differently: it is clamped down to the job's interval instead. `metrics-scraper` is the only job under 30s, and it resolves to a 15s timeout on its own — the value it would have been given explicitly.

Dropping the `blackbox` timeout changes nothing either: the prober uses the lower of its module timeout and the scrape timeout, so the module's 5s keeps governing the probe.

#### Resulting volume

| | Before | After |
|---|---|---|
| Scrapes/day (all targets) | 302,400 | **13,536** |
| earn-bot scrapes/day | 43,200 | **1,440** |
| balance-exporter RPC calls/day | ~389,000 | ~2,600 |

A **22x** reduction overall, and every target is still sampled at least twice as often as its data changes.

Dashboards and alert rules are untouched.

### Testing

`promtool check config` (v3.2.1) against the `prometheus.yml` key extracted from each configmap, plus staging's referenced `alert_rules.yml`:

```
 SUCCESS: prod/prometheus.yml is valid prometheus config file syntax
   SUCCESS: 1 rule files found
 SUCCESS: staging/prometheus.yml is valid prometheus config file syntax
   SUCCESS: 1 rules found
```

To confirm the check exercises the interval/timeout constraint rather than passing vacuously, a variant with an explicit `scrape_timeout: 90s` on a 60s job was rejected as expected:

```
FAILED: scrape timeout greater than scrape interval for scrape config with job name "earn-bot"
```

Since the clamping behaviour above is load-time rather than parse-time, `promtool` alone can't confirm it. The production config was loaded into a real Prometheus and the effective per-target values read back from `/api/v1/targets`:

```
targets: 7
  api-metrics        interval=1m    timeout=30s
  balance-exporter   interval=5m    timeout=30s
  balance-exporter   interval=5m    timeout=30s
  blackbox           interval=1m    timeout=30s
  bridge-worker      interval=30s   timeout=30s
  earn-bot           interval=1m    timeout=30s
  metrics-scraper    interval=15s   timeout=15s
```

Every job resolves to its intended tier, `metrics-scraper` clamps to 15s as expected, and no job silently inherits an unintended cadence.

Post-deploy this should show up as the intervals on Prometheus `/targets`, a drop to ~1.4k/day for earn-bot in Cloudflare analytics, and the end of `up` flapping on the earn-bot panels.

---

### Author's checklist

### Reviewer's checklist


lukasz-zimnoch (dismissed):

lukasz-zimnoch:
@nkuba commit is not signed

nkuba:
Let's update CLAUDE.md to enforce signing mezo-org/mezod#733

lukasz-zimnoch (approved):

PR #733, opened 2026-08-19
Make the commit signing rule unconditional

nkuba:
References: 

### Introduction

The signing convention in `CLAUDE.md` read:

> Commits **must** be GPG-signed (CI enforces; signing in the sandbox won't work — sign outside).

The carve-out invites an agent to decide for itself that its environment is one where signing cannot work, and to bypass signing on that basis. That is exactly what happened in a recent session: the agent read the note, concluded it applied, and passed `--no-gpg-sign`. Signing was working fine at the time.

The "CI enforces" half is also untrue — nothing in `.github/workflows/` or `.pre-commit-config.yaml` checks signatures.

### Changes

States the rule without conditions: never bypass signing, and stop and ask if it isn't working. Asking is cheap, and it puts the call in front of someone who can see whether signing is genuinely broken.

Per review, no environment is named at all. Naming one keeps the escape hatch available regardless of which one is named, and the failure this PR came from was in an environment the old note didn't even mention.

The false "CI enforces" claim is dropped rather than restated. If signing should actually be enforced, that belongs in a branch-protection ruleset — worth a follow-up, and viable, since remote agent sessions do produce properly signed commits.

### Testing

Documentation only. Lines stay within the configured `MD013` limit of 120 characters.

---

### Author's checklist

### Reviewer's checklist


CLAUDE.md
```diff
@@ -169,7 +169,10 @@ Three separate Hardhat projects — don't cross-import:
 
 ## Conventions
 
-- Commits **must** be GPG-signed (CI enforces; signing in the sandbox won't work — sign outside).
```
lukasz-zimnoch, line 172: Note that sandbox here refers to the built-in Claude sandbox mode (`/sandbox`) - not an external environment. This sentence is basically true but lacks some details. That said, I would not mention anything about remote environments and SSH signing (as we don't use that actually) but rather focus on expanding and disambiguating what the existing sentence actually means.

Shadowfiend: Really we should just tell it to stop and ask if signing isn't working 🤷

nkuba: We actually use remote environments and SSH signing in Claude Code remote sessions.
Agree, we don't need to be too specific in the instruction here, I updated it to be more general.

lukasz-zimnoch (approved):

